PR #1, opened 2018-02-16
Add steller resources

kbacha:


stellar_resources/src/utils.rs
```diff
@@ -0,0 +1,26 @@
+use serde::{Deserialize, Deserializer};
+use serde::de;
+use std::str::FromStr;
+
+pub fn deserialize_from_str<'de, D, T>(d: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: FromStr,
```
correlator: glad I made it through traits today.

stellar_resources/src/utils.rs
```diff
@@ -0,0 +1,26 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::value::Value;
+
+    #[test]
+    fn it_parses_a_string_to_u64() {
+        let value = Value::String("123".to_string());
+        assert_eq!(deserialize_from_str::<Value, u64>(value).unwrap(), 123u64);
+    }
```
correlator: What policy do we want to have on testing the error side of results?  I'm happy to go with the approach of letting the compiler enforce the error type and letting the calling functions worry about testing outcomes.  Shall we make this the agreed upon workflow?

stellar_resources/src/utils.rs
```diff
@@ -0,0 +1,26 @@
+use serde::{Deserialize, Deserializer};
+use serde::de;
+use std::str::FromStr;
+
+pub fn deserialize_from_str<'de, D, T>(d: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
```
correlator: Why does this require a lifetime?

kbacha: I think so that the return value has the same lifetime as the caller.

stellar_resources/src/utils.rs
kbacha: I was thinking about adding a failure case as well (like if `"abc"` comes in, since technically that "could" happen and we should make sure the outcome is correct (aka, doesn't panic).

I'd rather put the tests here than associated with the derived json struct.

stellar_resources/src/utils.rs
correlator: Interesting.  It catches me a bit off guard as it's being passed in as an argument and so I would expect it to have that scope out of the box but hey the compiler doesn't lie.  Let's ship this beast.

stellar_resources/src/utils.rs
correlator: I leave it up to you.

stellar_resources/src/utils.rs
kbacha: I can try it without it, the type signature from the documentation required it.

stellar_resources/src/utils.rs
kbacha: There is some logic, y'know? Like if it fails we return a particular error. I think it's probably important to include that as a test, and only test it here instead of all the places we include the derive. I'm adding a spec for it now.

stellar_resources/src/utils.rs
kbacha: Ok, so the issue is that the `Deserializer` trait takes a life time parameter, there for it's needed. It must contain a reference internally.

kbacha: https://github.com/serde-rs/serde/blob/master/serde/src/de/mod.rs#L497

So the trait requires it so it must be declared in the function signature. I think it's because it wasn't to be able to do "in place" deserialization, which would require keeping the references bound to the same lifetime.

PR #2, opened 2018-02-18
Implement basic client

kbacha:
In order to start moving forward, I've added a basic client. In doing
so, I got tired of typing `stellar_` over and over so simplified the
root directories. I also dropped the outer resource module since the
crate itself becomes a resource.

There's still a lot to do but this is the first building blocks. For end points, I want to return futures to the result but I want the namespacing to be part of the client:

```rust
let client = Client::new(/** */);
client.accounts()
  .get("id") // returns Future<Account> 
  .and_then(|account| /** do something */);
```


Cargo.toml
```diff
@@ -1,2 +1,2 @@
 [workspace]
-members = ["stellar_resources"]
+members = ["resources", "client"]
```
correlator: 👍

client/src/client/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+    /// # extern crate tokio_core;
+    /// # extern crate stellar_client;
+    /// # fn main() {
+    /// use tokio_core::reactor::Core;
+    /// use stellar_client::Client;
+    /// let core = Core::new().unwrap();
+    /// let client = Client::new("https://horizon-testnet.stellar.org", &core.handle());
+    /// # }
+    /// ```
+    pub fn new(uri: &str, handle: &Handle) -> Result<Self> {
```
correlator: URI feels more like a configuration to me than a parameter, but I'm happy to come back to that later.

kbacha: I'm not sure what you mean?

In order to create an instance of the client you need to hand it a uri in some form. This could be through a builder pattern but since it's just 2 params at the moment I felt a param made more sense. It's how you "configure" the client.

PR #3, opened 2018-02-20
Create convenience methods around client construction

kbacha:
To make it easier to use the horizon and horizon test end points
convenience methods have been added to make that easy. In addition,
tests around the construction (along with documentation) have been
written.

### DIFF
--- a/client/src/client/mod.rs
+++ b/client/src/client/mod.rs
@@ -3,12 +3,22 @@ use hyper_tls::HttpsConnector;
 use tokio_core::reactor::Handle;
 use error::{Result, Error};
 
+#[derive(Debug, Clone, PartialEq)]
+enum Host {
+    HorizonTest,
+    HorizonProd,
+    Other(String),
+}
+
 /// A client that can issue requests to a horizon api.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Client {
     inner: hyper::Client<HttpsConnector<hyper::client::HttpConnector>>,
-    uri: hyper::Uri,
+    host: Host,
 }
+
+static HORIZON_TEST_URI: &'static str = "https://horizon-testnet.stellar.org";
+static HORIZON_URI: &'static str = "https://horizon.stellar.org";
 
 impl Client {
     /// Constructs a new stellar client.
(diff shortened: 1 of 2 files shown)


correlator:
🚢

client/src/client/mod.rs
```diff
@@ -3,13 +3,23 @@ use hyper_tls::HttpsConnector;
 use tokio_core::reactor::Handle;
 use error::{Result, Error};
 
+#[derive(Debug, Clone, PartialEq)]
+enum Host {
+    HorizonTest,
+    HorizonProd,
+    Other(String),
```
correlator: 💪

client/src/client/mod.rs
```diff
@@ -22,14 +32,142 @@ impl Client {
...
     /// use stellar_client::Client;
     /// let core = Core::new().unwrap();
-    /// let client = Client::new("https://horizon-testnet.stellar.org", &core.handle());
+    /// let client = Client::new("https://horizon-testnet.stellar.org", &core.handle()).unwrap();
     /// # }
     /// ```
     pub fn new(uri: &str, handle: &Handle) -> Result<Self> {
+        // Ensure that the uri passed in can parse.
+        let _: hyper::Uri = uri.parse().map_err(|_| Error::BadUri)?;
+        Self::build(Host::Other(uri.to_string()), &handle)
```
correlator: :)

client/src/client/mod.rs
```diff
@@ -22,14 +32,142 @@ impl Client {
...
+    }
+
+    fn build(host: Host, handle: &Handle) -> Result<Self> {
```
correlator: would it make sense to have this be a public method in a future PR?

client/src/client/mod.rs
```diff
@@ -22,14 +32,142 @@ impl Client {
...
+    /// # extern crate stellar_client;
+    /// # fn main() {
+    /// use tokio_core::reactor::Core;
+    /// use stellar_client::Client;
+    /// let core = Core::new().unwrap();
+    /// let client = Client::horizon_test(&core.handle()).unwrap();
+    /// # }
+    /// ```
+    pub fn horizon_test(handle: &Handle) -> Result<Self> {
+        Self::build(Host::HorizonTest, &handle)
```
correlator: 😎

correlator: 🔥

client/src/client/mod.rs
correlator: Nevermind.  I see what you did below.  This is 🌶

PR #5, opened 2018-02-20
add assets to resources

correlator:
@kbacha 

There was some funny business that needed to be handled in the amount field.  Do you think there is anything else we should add to call out this behavior?

### DIFF
--- a/resources/src/asset.rs
+++ b/resources/src/asset.rs
@@ -1,0 +1,99 @@
+use utils;
+/// Permissions around who can own an asset and whether or
+/// not the asset issuer can freeze the asset.
+#[derive(Serialize, Deserialize, Debug)]
+struct Flag {
+    auth_required: bool,
+    auth_revocable: bool,
+}
+
+/// Assets are the units that are traded on the Stellar Network.
+/// An asset consists of an type, code, and issuer.
+/// Any asset can be traded for any other asset.
+/// https://www.stellar.org/developers/horizon/reference/resources/asset.html
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Asset {
+    asset_type: String,
+    asset_code: String,
+    asset_issuer: String,
+    #[serde(deserialize_with = "utils::deserialize_from_str")] amount: f64,
+    num_accounts: u32,
+    flags: Flag,
+}
+
+impl Asset {
+    ///   The type of this asset: “credit_alphanum4”, or “credit_alphanum12”.
+    ///   Returns a slice that lives as long as the asset does.
+    pub fn asset_type<'a>(&'a self) -> &'a str {
+        &(self.asset_type)
+    }
+
+    /// The code of this asset.
+    /// Returns a slice that lives as long as the asset does.
+    pub fn asset_code<'a>(&'a self) -> &'a str {
+        &(self.asset_code)
+    }
+
+    /// The issuer of this asset.  This corresponds to the id of an account.
+    /// Returns a slice that lives as long as the asset does.
+    pub fn asset_issuer<'a>(&'a self) -> &'a str {
+        &(self.asset_issuer)
+    }
+
+    /// The number of units of credit issued for this asset.
+    /// This number is scaled by 10 million to display the number if the format a
+    /// user would expect it in
+    /// https://www.stellar.org/developers/guides/concepts/assets.html
+    /// Returns a signed 64-bit integer.
+    pub fn amount(&self) -> i64 {
+        (self.amount * 10_000_000.0) as i64
+    }
+
+    /// The number of accounts that: 1) trust this asset and 2) where if the asset has the
+    /// auth_required flag then the account is authorized to hold the asset.
+    /// Returns an unsigned 32-bit integer
+    pub fn num_accounts(&self) -> u32 {
+        self.num_accounts
+    }
+
+    /// If this field is true it means the anchor must approve anyone who wants to
+    /// hold its credit, allowing it to control who its customers are
+    /// Returns a bool.
+    pub fn auth_required(&self) -> bool {
+        self.flags.auth_required
+    }
+
+    /// If this field is true it means the anchor can freeze credit held by another account. When
+    /// credit is frozen for a particular account, that account can only send the credit back to
+    /// the anchor–it can’t transfer the credit to any other account. This setting allows the
+    /// issuing account to revoke credit that it accidentally issued or that was obtained
+    /// improperly.
+    /// Returns a bool.
+    pub fn auth_revocable(&self) -> bool {
+        self.flags.auth_revocable
+    }
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json;
+
+    fn asset_json() -> &'static str {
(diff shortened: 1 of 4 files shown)


resources/src/asset.rs
```diff
@@ -0,0 +1,99 @@
...
+        &(self.asset_issuer)
+    }
+
+    /// The number of units of credit issued for this asset.
+    /// This number is scaled by 10 million to display the number if the format a
+    /// user would expect it in
+    /// https://www.stellar.org/developers/guides/concepts/assets.html
+    /// Returns a signed 64-bit integer.
+    pub fn amount(&self) -> i64 {
+        (self.amount * 10_000_000.0) as i64
```
kbacha: Two things come to mind here:

1. We don't want to incur a multiplication every time we look at amount.
2. Multiplying large floating points can lead to loss of precision which would not be good here.

I'd recommend implementing a custom deserializer to convert to amount. One where we determine the offset of the decimal point, sub out the decimal from a string, then convert to a i64, and multiply by the orders of maginitude between the offset at 10 million.

Definitely slightly tricky code but the result should retain the precision that stellar has.

resources/src/asset.rs
```diff
@@ -0,0 +1,99 @@
...
+    asset_type: String,
+    asset_code: String,
+    asset_issuer: String,
+    #[serde(deserialize_with = "utils::deserialize_from_str")] amount: f64,
+    num_accounts: u32,
+    flags: Flag,
+}
+
+impl Asset {
+    ///   The type of this asset: “credit_alphanum4”, or “credit_alphanum12”.
```
kbacha: Extra leading spaces in the comment.

resources/src/asset.rs
```diff
@@ -0,0 +1,99 @@
...
+    ///   Returns a slice that lives as long as the asset does.
+    pub fn asset_type<'a>(&'a self) -> &'a str {
+        &(self.asset_type)
```
kbacha: Do you need the surrounding parens for all these?

resources/src/asset.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    #[test]
+    fn it_parses() {
+        let asset: Asset = serde_json::from_str(&asset_json()).unwrap();
+        assert_eq!(asset.asset_type(), "credit_alphanum4");
+        assert_eq!(asset.asset_code(), "USD");
+        assert_eq!(asset.asset_issuer(), "GBAUUA74H4XOQYRSOW2RZUA4QL5PB37U3JS5NE3RTB2ELJVMIF5RLMAG");
+        assert_eq!(asset.amount(), 1000000000);
+        assert_eq!(asset.num_accounts(), 91547871);
+        assert_eq!(asset.auth_required(), false);
```
kbacha: For booleans you can use:
```rust
assert!(!asset.auth_required());
```

resources/src/asset.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// auth_required flag then the account is authorized to hold the asset.
+    /// Returns an unsigned 32-bit integer
+    pub fn num_accounts(&self) -> u32 {
+        self.num_accounts
+    }
+
+    /// If this field is true it means the anchor must approve anyone who wants to
+    /// hold its credit, allowing it to control who its customers are
+    /// Returns a bool.
+    pub fn auth_required(&self) -> bool {
```
kbacha: I like starting with a helping verb in rust since `?` is not valid in an identifier:

```rust
pub fn is_auth_required(&self) -> bool {
```

resources/src/asset.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Assets are the units that are traded on the Stellar Network.
+/// An asset consists of an type, code, and issuer.
+/// Any asset can be traded for any other asset.
+/// https://www.stellar.org/developers/horizon/reference/resources/asset.html
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Asset {
+    asset_type: String,
+    asset_code: String,
+    asset_issuer: String,
+    #[serde(deserialize_with = "utils::deserialize_from_str")] amount: f64,
```
kbacha: ugh, i just noticed that rustfmt does this. I think i prefer them on the previous lines, we may want to check in a rustfmt yml for config at some point...  #6 (just created)

resources/src/asset.rs
```diff
@@ -0,0 +1,117 @@
...
+                let decimal_place = self.amount.rfind('.').unwrap();
+                // Extra -1 to account for indexing differences
+                let number_decimals = length - 1 - decimal_place;
+                let mut new_amount = self.amount.clone();
+                // remove the decimal
+                let new_amount = new_amount.replace(".", "");
+                // Stellar sends a float that is reduced from true value by 10^7 so raise by 10 
+                // minus the amount we gained from removing decimal
+                let required_power: u32 = (7 - number_decimals) as u32;
+                return (new_amount.parse::<i64>().unwrap()) * 10_i64.pow(required_power);
```
correlator: I would love to find a way to set the optional value within this None arm, but couldn't find anything and going to be.  @kbacha maybe you've got a nice idea of how to handle this.

resources/src/asset.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /// The number of units of credit issued for this asset.
+    /// This number is scaled by 10 million to display the number if the format a
+    /// user would expect it in
+    /// https://www.stellar.org/developers/guides/concepts/assets.html
+    /// Returns a signed 64-bit integer.
+    pub fn amount(&self) -> i64 {
+        match self.formatted_amount {
+            Some(formatted_amount) => return formatted_amount,
```
kbacha: This appears to never be possible, correct?

resources/src/asset.rs
```diff
@@ -0,0 +1,117 @@
...
+            None => {
+                // A string with 1 char has len() == 1
+                let length = self.amount.len();
+                // A string starting with . has rfind('.') == 0
```
kbacha: Based on the documentation it looks like this would actually return `None` not `0`
https://play.rust-lang.org/?gist=a2c0d7d6b66f7861536c14faab208043&version=stable

resources/src/asset.rs
```diff
@@ -0,0 +1,117 @@
...
+                let decimal_place = self.amount.rfind('.').unwrap();
```
kbacha: If you want to unwrap here, try using `.unwrap_or(length)` so that if it doesn't find anything you default to the length of the string. If you want it to be 0 then you can use 0.

resources/src/asset.rs
```diff
@@ -0,0 +1,117 @@
...
+                // Extra -1 to account for indexing differences
+                let number_decimals = length - 1 - decimal_place;
+                let mut new_amount = self.amount.clone();
```
kbacha: This is not needed as `replace` creates a new string. You actually end up copying the data twice.

resources/src/asset.rs
kbacha: You cannot mutate a non-mut self reference. We should tackle the deserialize before the struct is made.

correlator:
@kbacha

PR #20, opened 2018-02-24
Add offer

kbacha:
Several changes:

1. Refactored amounts and other deserializers into their own module for just that purpose
2. Refactored the amount deserializer to be slightly more efficient
3. Added offers and partial asset types

See commit descriptions for more details.


resources/src/account.rs
```diff
@@ -1,4 +1,4 @@
-use utils;
+use deserialize;
```
correlator: 👍

resources/src/asset.rs
```diff
@@ -1,6 +1,37 @@
...
+use deserialize;
+
+/// Assets are the units that are traded on the Stellar Network.
+/// An asset consists of an type, code, and issuer.
+/// Any asset can be traded for any other asset.
+/// https://www.stellar.org/developers/horizon/reference/resources/asset.html
+///
+/// A partial is just the type, code, issuer
+#[derive(Serialize, Deserialize, Debug)]
+pub struct PartialAsset {
```
correlator: I like this move especially since this is used so frequently in all of the offer payloads.  I might suggest calling this AssetIdentifier seeing as these are the three fields required to uniquely identify an asset in Stellar.  What are your thoughts?

resources/src/deserialize.rs
```diff
@@ -0,0 +1,93 @@
+use serde::{Deserialize, Deserializer};
+use serde::de;
+use std::str::FromStr;
+
+/// The stellar ammount fields are represented as strings in the horizon api,
+/// however, they look like floats. But, in the actual ledger they are signed
+/// 64-bit integers. This function can be used to convert what we get from
```
correlator: switch "can be used to convert" to "converts"

resources/src/deserialize.rs
```diff
@@ -0,0 +1,93 @@
...
+/// horizon into a signed integer. The precision of the float is assumed to
+/// by out to 7 digits (but if fewer are found it's ok).
```
correlator: assumed to "be"

resources/src/deserialize.rs
```diff
@@ -0,0 +1,93 @@
...
+pub(crate) fn amount<'de, D>(d: D) -> Result<i64, D::Error>
```
correlator: Didn't know you could do this `pub(crate)`  awesome!

resources/src/deserialize.rs
```diff
@@ -0,0 +1,93 @@
...
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(d)?;
+    match s.rfind('.') {
+        None => {
```
correlator: 😎

resources/src/deserialize.rs
```diff
@@ -0,0 +1,93 @@
...
+        Some(decimal_place) => {
+            let number_decimals = s.len() - (decimal_place + 1);
+            if number_decimals > 7 {
+                Err(de::Error::custom(
+                    "Amount has too many digits of precision.",
+                ))
+            } else {
+                let s = s.replace(".", "");
+                let parsed_amount = i64::from_str(&s)
+                    .map_err(|_| de::Error::custom("Failed to parse string field"))?;
```
correlator: Both arms compute the parsed amount.  Can we do this before the match?

resources/src/deserialize.rs
```diff
@@ -0,0 +1,93 @@
...
+    #[test]
+    fn it_errors_floats_with_more_than_7_decimals() {
+        let value = Value::String("0.212847948".to_string());
+        assert!(super::amount(value).is_err());
+    }
+}
+
+/// Some fields in the json are represented as "strings" but are actually
+/// another type. If that type implements the `FromStr` trait, then this
+/// can properly parse it into that type.
```
correlator: again I personally prefer the active voice "this properly parses it into that type"

resources/src/offer.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// An offer being made for particular assets at a particular exchange rate.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Offer {
+    id: i64,
+    paging_token: String,
+    seller: String,
+    selling: PartialAsset,
+    buying: PartialAsset,
```
correlator: THIS
IS
🔥

resources/src/offer.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[serde(deserialize_with = "deserialize::amount")] amount: i64,
+    #[serde(rename = "price_r")] price_ratio: PriceRatio,
+    #[serde(deserialize_with = "deserialize::amount")] price: i64,
```
correlator: I think both price and amount should be floats otherwise you could only trade whole units of token

resources/src/asset.rs
kbacha: Good idea, I like `AssetIdentifier`, I was having trouble coming up with a good name.

resources/src/deserialize.rs
kbacha: Yes, i considered that, but this is actually more efficient. See, on this arm it won't executes after we've removed the decimal point and after we've checked the precision.

correlator: 👍

resources/src/asset.rs
```diff
@@ -1,6 +1,37 @@
...
-use serde::de;
-use std::str::FromStr;
+use deserialize;
+
+/// Assets are the units that are traded on the Stellar Network.
+/// An asset consists of an type, code, and issuer.
+/// Any asset can be traded for any other asset.
+/// https://www.stellar.org/developers/horizon/reference/resources/asset.html
+///
+/// A partial is just the type, code, issuer
```
correlator: An Identifier

resources/src/asset.rs
```diff
@@ -148,4 +131,15 @@ mod asset_tests {
         assert!(!asset.is_auth_required());
         assert!(asset.is_auth_revocable());
     }
+
+    #[test]
+    fn it_parses_a_partial() {
```
correlator: an_identifier

PR #22, opened 2018-02-25
Scott/amounts

### LINKED ISSUE #19: Implement a "type" for amount
Turn the i64 we've been using into a tuple struct that knows how to implement Serde but also knows how to display itself

```rust
struct Amount(i64);
impl Display for Amount {
}

impl Serialize for Amount {
}

impl Deserialize for Amount {
}
```

Traits good for implementing:

- [x] Display https://doc.rust-lang.org/std/fmt/trait.Display.html
- [x] Serialize https://docs.serde.rs/serde/ser/trait.Serialize.html (should allow us to write to json)
- [x] Deserialize https://docs.serde.rs/serde/de/trait.Deserialize.html (should allow us to read from json)
- [x] Add (and subtract and others) https://doc.rust-lang.org/std/ops/trait.Add.html
- [x] PartialEq and Eq (might be derivable) https://doc.rust-lang.org/std/cmp/trait.Eq.html
- [x] PartialOrd and Ord https://doc.rust-lang.org/std/cmp/trait.Ord.html https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html

Some tests might be:

```rust
fn it_displays_as_float() {
  assert_eq!(format!(Amount(10)), "0.0000010");
}

fn it_orders_itself() {
    assert!(Amount(10) < Amount(100));
}

fn it_equals_itself() {
    assert_eq!(Amount(10), Amount(10));
    assert_ne!(Amount(100), Amount(10));
}
```

correlator:
@kbacha 

Creates new amount resource

- implements serialize and deserialize traits
- implements add subtract, compare and copy traits
- Orders and Assets now use Amounts instead of i64 as field types

resolves #19


resources/src/amount.rs
```diff
@@ -0,0 +1,192 @@
+use std::fmt;
+use std::ops::{Add, Sub};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::de;
+use std::str::FromStr;
+
+/// Amounts are used in serveral resources in the stellar ecosystem.  There
```
kbacha: typo: serveral

kbacha:
I was wondering if we should call it `AssetAmount` instead of just amount. But i'm cool either way.

PR #23, opened 2018-02-26
Scott/orderbook

### LINKED ISSUE #16: Add Order Book
The collection of offers available to make exchanges for.

### LINKED ISSUE #21: Assert that  amounts cannot be negative
Amounts are always positive (even though they are represented by a signed integer). We should protect against bad amount data.

### LINKED ISSUE #24: Assets should only default asset_code to XLM if asset_type is "native"

correlator:
@kbacha 
This branch does the following 3 things:

- Add the order book resource resolves #16
- Set XLM as the default asset (Opening an issue to make this only be the case when type is "native") resolves #24 
- Amounts must be positive resolves #21


resources/src/amount.rs
```diff
@@ -15,6 +15,10 @@ pub struct Amount(i64);
 impl Amount {
     /// Create an amount newtype
     pub fn new(amount: i64) -> Amount {
+        assert!(
+            amount >= 0,
+            format!("Amount must be greater than zero, got {}.", amount)
```
kbacha: this will always create the string even if we aren't panicking. What does the err say if you don't put the message in?

resources/src/asset.rs
```diff
@@ -1,18 +1,60 @@
...
+            .map_err(|_| de::Error::custom("Code and issuer are required for non-native assets"))
+    }
+}
+
+impl Serialize for AssetIdentifier {
+    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        if self.asset_type == "native".to_string() {
```
kbacha:
```rust
let rep = if self.is_native() { // perhaps switch to a method?
  IntermediateAssetIdentifier::native()  // Add a method to return the native?
} else {
IntermediateAssetIdentifier {
  asset_type: self.asset_type.to_owned(),
  asset_code: Some(self.asset_code.to_owned()),
  asset_issuer: Some(self.asset_issuer.to_owned()),
}
rep.serialize(s)
```

I think its better to have a single return here.

resources/src/asset.rs
```diff
@@ -31,11 +73,93 @@ impl AssetIdentifier {
...
     }
+
+    /// A new Asset can be a native stellar, or a fully identified asset
+    pub fn new(
+        asset_type: String,
+        asset_code: Option<String>,
+        asset_issuer: Option<String>,
+    ) -> Result<AssetIdentifier, String> {
+        if asset_type == "native".to_string() {
+            Ok(AssetIdentifier {
```
kbacha: This is always the same, could just be a method.

resources/src/asset.rs
```diff
@@ -31,11 +73,93 @@ impl AssetIdentifier {
...
         &self.asset_issuer
     }
+
+    /// A new Asset can be a native stellar, or a fully identified asset
+    pub fn new(
+        asset_type: String,
+        asset_code: Option<String>,
+        asset_issuer: Option<String>,
+    ) -> Result<AssetIdentifier, String> {
+        if asset_type == "native".to_string() {
```
kbacha: What happened to making the asset type an enum?

resources/src/asset.rs
```diff
@@ -31,11 +73,93 @@ impl AssetIdentifier {
...
+            Ok(AssetIdentifier {
+                asset_type: asset_type,
+                asset_code: "XLM".to_string(),
+                asset_issuer: "Stellar Foundation".to_string(),
+            })
+        } else {
+            if asset_code.is_none() || asset_issuer.is_none() {
+                Err("Code and issuer are required for non-native assets".to_owned())
```
kbacha: String errs are less ideal, we should implement an enum for this or return a unit `Err(())` since we map away the error anyhow.

This is the equivalent of `raise "some string"` in ruby and is frowned upon.

resources/src/asset.rs
```diff
@@ -45,33 +169,89 @@ struct Flag {
...
 
+impl<'de> Deserialize<'de> for Asset {
+    fn deserialize<D>(d: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let rep: IntermediateAsset = IntermediateAsset::deserialize(d)?;
+        let asset_identifier: Result<AssetIdentifier, D::Error> =
+            AssetIdentifier::new(rep.asset_type, rep.asset_code, rep.asset_issuer).map_err(|_| {
+                de::Error::custom("Code and issuer are required for non-native assets")
```
kbacha: This assumes to know what the error is when it could be any `String`. We should probably match on an explicit error type.

resources/src/asset.rs
```diff
@@ -45,33 +169,89 @@ struct Flag {
...
+        if self.asset_type() == "native" {
+            let rep = IntermediateAsset {
+                asset_type: self.asset_type().to_owned(),
+                asset_code: None,
+                asset_issuer: None,
+                amount: self.amount,
+                num_accounts: self.num_accounts,
+                flags: self.flags.clone(),
+            };
+            return rep.serialize(s);
```
kbacha: As with above, better to have a single return.

resources/src/offer.rs
```diff
@@ -8,6 +8,21 @@ pub struct PriceRatio {
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct OfferSummary {
+    amount: Amount,
+    #[serde(rename = "price_r")] price_ratio: PriceRatio,
+    price: Amount,
+}
+
+impl OfferSummary {
+    /// Returns the amount of an asset the offer is willing to buy or sell
+    pub fn amount(&self) -> Amount {
```
kbacha: Test?

resources/src/offer.rs
```diff
@@ -8,6 +8,21 @@ pub struct PriceRatio {
     #[serde(rename = "d")] denominator: u64,
 }
 
+/// Summary of an offer to be shown in an orderbook
+#[derive(Serialize, Deserialize, Debug)]
+pub struct OfferSummary {
+    amount: Amount,
+    #[serde(rename = "price_r")] price_ratio: PriceRatio,
+    price: Amount,
```
kbacha: Shouldn't we implement the other public methods of this?

resources/src/asset.rs
correlator: Lets address this in a separate pr :)

resources/src/asset.rs
```diff
@@ -1,41 +1,181 @@
...
+    /// Asset with a 12 character code
+    CreditAlphanum12(AssetId),
+}
+
+/// Struct containing asset_code and asset_issuer
 #[derive(Serialize, Deserialize, Debug)]
-pub struct AssetIdentifier {
-    asset_type: String,
+pub struct AssetId {
     asset_code: String,
```
kbacha: optionally remove superfluous `asset_` prefixes

PR #26, opened 2018-02-27
Add the trade resource.

### LINKED ISSUE #15: Add trade
Represents a fulfulled order

https://www.stellar.org/developers/horizon/reference/resources/trade.html

kbacha:
This adds the trade resource to the repo and implements it with a
intermediate resource that represents the actual json structure. This
allows easy derivation of serde but still only returns a Trade struct
with the data we want.

resolves #15 

PS: To complete the asset identifier implementation it depends on #23

### DIFF
--- a/resources/src/offer.rs
+++ b/resources/src/offer.rs
@@ -2,10 +2,30 @@ use amount::Amount;
 use asset::AssetIdentifier;
 
 /// The ratio between the asking and selling price
-#[derive(Serialize, Deserialize, Debug, Copy, Clone)]
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Copy, Clone)]
 pub struct PriceRatio {
     #[serde(rename = "n")] numerator: u64,
     #[serde(rename = "d")] denominator: u64,
+}
+
+impl PriceRatio {
+    /// Constructs a new price ratio struct
+    pub fn new(numerator: u64, denominator: u64) -> Self {
+        Self {
+            numerator,
+            denominator,
+        }
+    }
+
+    /// The numerator
+    pub fn numerator(&self) -> u64 {
+        self.numerator
+    }
+
+    /// The denominator
+    pub fn denominator(&self) -> u64 {
+        self.denominator
+    }
 }
 
 /// Summary of an offer to be shown in an orderbook
--- a/resources/src/trade.rs
+++ b/resources/src/trade.rs
@@ -1,0 +1,301 @@
+use chrono::prelude::*;
+use amount::Amount;
+use asset::AssetIdentifier;
+use offer::PriceRatio;
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+
+/// A trade represents an offer that was fulfilled between two assets and accounts.
+///
+/// https://www.stellar.org/developers/horizon/reference/resources/trade.html
+#[derive(Debug)]
+pub struct Trade {
+    id: String,
+    offer_id: String,
+    paging_token: String,
+    ledger_close_time: DateTime<Utc>,
+    base_account: String,
+    base_amount: Amount,
+    base_asset: AssetIdentifier,
+    counter_amount: Amount,
+    counter_account: String,
+    counter_asset: AssetIdentifier,
+    price: PriceRatio,
+    seller: Seller,
+}
+
+/// The seller involved in the trade.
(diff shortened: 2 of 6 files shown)


resources/src/trade.rs
```diff
@@ -0,0 +1,253 @@
...
+/// https://www.stellar.org/developers/horizon/reference/resources/trade.html
+#[derive(Debug)]
+pub struct Trade {
+    id: String,
+    offer_id: String,
+    paging_token: String,
+    ledger_close_time: DateTime<Utc>,
+    base_account: String,
+    base_amount: Amount,
+    // base_asset: AssetIdentifier,
```
correlator: ?

resources/src/trade.rs
```diff
@@ -0,0 +1,253 @@
...
+        *self == Seller::Counter
+    }
+}
+
+impl<'de> Deserialize<'de> for Trade {
+    fn deserialize<D>(d: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let rep: TradeRepresentation = TradeRepresentation::deserialize(d)?;
```
correlator: WOW

resources/src/trade.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "UPPERCASE")]
+struct Price {
+    n: u64,
+    d: u64,
+}
+
+impl From<Price> for PriceRatio {
```
correlator: Should this be in the offer file where PriceRatio is defined?  Not sure of the convention here

resources/src/trade.rs
```diff
@@ -0,0 +1,253 @@
...
+        let rep = TradeRepresentation {
+            id: self.id.to_owned(),
+            paging_token: self.paging_token.to_owned(),
+            ledger_close_time: self.ledger_close_time,
+            offer_id: self.offer_id.to_owned(),
+            base_account: self.base_account.to_owned(),
+            base_amount: self.base_amount,
+            base_asset_type: "".to_string(),
+            base_asset_code: None,
+            base_asset_issuer: None,
```
correlator: Do we do this because all the asset information should be contained in the offer?

resources/src/trade.rs
```diff
@@ -0,0 +1,253 @@
...
+    id: String,
+    offer_id: String,
+    paging_token: String,
+    ledger_close_time: DateTime<Utc>,
+    base_account: String,
+    base_amount: Amount,
+    // base_asset: AssetIdentifier,
+    counter_amount: Amount,
+    counter_account: String,
+    // counter_asset: AssetIdentifier,
```
correlator: ?

resources/src/trade.rs
```diff
@@ -0,0 +1,253 @@
+use chrono::prelude::*;
+use amount::Amount;
+// use asset::AssetIdentifier;
```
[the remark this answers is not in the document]
kbacha: Was leaving in for when you finish your pr ;p

resources/src/trade.rs
kbacha: The assets have identifiers in the trade, i'm waiting on the "native" support in your branch.

resources/src/trade.rs
kbacha: No, the 'Price' here is private to the trade, so it should only be here.

resources/src/trade.rs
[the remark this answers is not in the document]
correlator: I suspected as much ;)

resources/src/offer.rs
```diff
@@ -2,12 +2,32 @@ use amount::Amount;
...
+impl PriceRatio {
+    /// Constructs a new price ratio struct
+    pub fn new(numerator: u64, denominator: u64) -> Self {
+        Self {
+            numerator,
+            denominator,
+        }
+    }
+
+    /// The numerator
```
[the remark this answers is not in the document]
kbacha: Yea... once i figure out what it means.

resources/src/trade.rs
correlator: I used the keyword "intermediate" and you are using the keyword "representation" I think both are fine, "Denormalized" might be more accurate than both, but we should pick one and stick with it.  We can also address this in a future pr.

PR #29, opened 2018-03-02
add resource for stellar errors in the Horizon API

### LINKED ISSUE #9: Add errors
There are a bunch of types of errors (which can likely be captured in an enum). But the main format appears to be the same for all:
https://www.stellar.org/developers/horizon/reference/errors.html

This may be more apt to be modeled in the client library or as a separate crate.

correlator:
Add Stellar Horizon API errors resolves #9 

I thought about putting these into the client, but that would require pulling serde into the client.  They also implement the "Problem Details for HTTP APIs" standard which makes error messages behave a lot like the other things we serialize and deserialize in our resources library.


resources/src/stellar_error.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// A resource for the stellar horizon API specific error codes.
+/// These errors adhere to the Problem Details Standard
+/// https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-00
+/// and a list of possible erros can be found here
+/// https://www.stellar.org/developers/horizon/reference/errors.html
+#[derive(Deserialize)]
+pub struct StellarError {
+    // type is a protected word
+    #[serde(rename = "type")] problem_type: String,
```
kbacha: Could this be enumerated?

resources/src/stellar_error.rs
```diff
@@ -0,0 +1,158 @@
...
+    use serde_json;
+
+    fn before_history_json() -> &'static str {
+        include_str!("../fixtures/before_history_error.json")
+    }
+
+    #[test]
+    fn it_parses_native_assets_from_json() {
+        let before_history: StellarError = serde_json::from_str(&before_history_json()).unwrap();
+        assert_eq!(println!("{}", before_history), "This horizon instance is configured to only track a portion of the stellar network's latest history. This request is asking for results prior to the recorded history known to this horizon instance.");
```
kbacha: Don't use `println!` to test formatting, use `format!`

resources/src/stellar_error.rs
```diff
@@ -0,0 +1,158 @@
...
+/// These errors adhere to the Problem Details Standard
+/// https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-00
+/// and a list of possible erros can be found here
+/// https://www.stellar.org/developers/horizon/reference/errors.html
+#[derive(Deserialize)]
+pub struct StellarError {
+    // type is a protected word
+    #[serde(rename = "type")] problem_type: String,
+    title: String,
+    status: u16,
```
kbacha: This aspect is tightly coupled to http and it would be nice if it mapped into the enumerated type for http status codes. Both Hyper and http crate have these as enumerations.

resources/src/stellar_error.rs
```diff
@@ -0,0 +1,158 @@
...
+    detail: String,
```
kbacha: I believe this may be optional
https://www.stellar.org/developers/horizon/reference/errors/not-acceptable.html

correlator:
@kbacha moved everything into the client.  hyper does not implement deserialize on its status codes and I don't feel like going through this by hand right now so keeping as u16

client/src/stellar_error.rs
```diff
@@ -0,0 +1,244 @@
...
+            "forbidden" => ErrorType::Forbidden,
+            "not_acceptable" => ErrorType::NotAcceptable,
+            "not_found" => ErrorType::NotFound,
+            "not_implemented" => ErrorType::NotImplemented,
+            "rate_limit_exceeded" => ErrorType::RateLimitExceeded,
+            "internal_server_error" => ErrorType::InternalServerError,
+            "stale_history" => ErrorType::StaleHistory,
+            "transaction_failed" => ErrorType::TransactionFailed,
+            "transaction_malformed" => ErrorType::TransactionMalformed,
+            _ => ErrorType::NotAValidErrorType,
```
kbacha: I think accepting this as an `Other(String)` might be better so that we can keep the value if they add an additional type

client/src/stellar_error.rs
```diff
@@ -0,0 +1,244 @@
...
+    StaleHistory,
+    TransactionFailed,
+    TransactionMalformed,
+    OtherError(String),
+}
+
+#[derive(Deserialize)]
+struct IntermediateStellarError {
+    // type is a protected word
+    #[serde(rename = "type")] error_type: String,
```
kbacha: If the only thing being serialized here is the type, then you probably could have just implemented serde for the enum and not needed an intermediate type.

```rust
impl<'de> Deserialize<'de> for ErrorType {
  fn deserialize<'de>(d: Deserializer) -> Result {
    match String::deserialize(d)? {
      "type" => EnumType
    }
}
```

client/src/stellar_error.rs
```diff
@@ -0,0 +1,244 @@
...
+            rep.status,
+            rep.detail,
+            rep.instance,
+        ).map_err(|err| de::Error::custom(err))
+    }
+}
+
+impl StellarError {
+    /// Creates a new stellar error.  If the error type is not recognized an error is returned.
+    pub fn new(
```
kbacha: What's the advantage of this being public?

client/src/stellar_error.rs
correlator: This was a great suggestion thank you.

client/src/stellar_error.rs
correlator: bad answer but I use it in a test :/ I don't think an external lib would ever have a reason to call this though.

PR #32, opened 2018-03-04
add payment paths

### LINKED ISSUE #12: Add payment path
The path to exchange for the assets requested:
https://www.stellar.org/developers/horizon/reference/resources/path.html

correlator:
Added payment paths to resources, resolves #12 

This required exposing a little more of the asset identifier resources as they are heavily used and passed around in serialization.


resources/src/payment_path.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn new(
+        path: Vec<AssetIdentifier>,
+        source_amount: Amount,
+        destination_amount: Amount,
+        destination_asset_type: String,
+        destination_asset_code: Option<String>,
+        destination_asset_issuer: Option<String>,
+        source_asset_type: String,
+        source_asset_code: Option<String>,
+        source_asset_issuer: Option<String>,
```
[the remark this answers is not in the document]
kbacha: I'd rather have `new` (since it's public) to just take two assets as opposed to 6 extra params.

resources/src/payment_path.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// Source asset specified in the search that found this path
+    pub fn source_asset<'a>(&'a self) -> &'a AssetIdentifier {
+        &self.source_asset
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct IntermediatePaymentPath {
```
kbacha: Why does this need to be public?

correlator: left over from when i was thinking about calling new with an IntermediatePaymentPath.  Thanks for catching

kbacha:
Something seems messed up about this PR.

PR #33, opened 2018-03-04
Define EndPoint trait for usage with client

kbacha:
In order to produce definitions for every end point, we need a generic
trait that defines the entry and exit. This is a first draft of what
that might look like.

It doesn't yet interact with the client.


client/src/endpoint/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+use hyper::Uri;
+use serde::Deserialize;
+
+mod account;
+pub use self::account::AccountDetails;
+
+pub(crate) trait EndPoint<'de> {
+    type Response: Deserialize<'de>;
+
+    fn to_uri(&self, host: &str) -> Result<Uri>;
```
correlator: 👍

client/src/endpoint/mod.rs
```diff
@@ -0,0 +1,14 @@
+//! This module contains the various end point definitions for stellar's horizong
```
correlator: extra g at end of line

PR #34, opened 2018-03-06
add payment path resource

### LINKED ISSUE #12: Add payment path

correlator:
Added payment paths to resources, resolves #12 

This required exposing a little more of the asset identifier resources as they are heavily used and passed around in serialization.

### DIFF
--- a/resources/src/asset.rs
+++ b/resources/src/asset.rs
@@ -7,7 +7,7 @@ use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 /// https://www.stellar.org/developers/horizon/reference/resources/asset.html
 
 /// An identifer is the type, code, and issuer.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub enum AssetIdentifier {
     /// Stellar Lumens!
     Native,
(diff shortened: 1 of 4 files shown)


resources/src/asset.rs
```diff
@@ -85,6 +85,15 @@ impl AssetIdentifier {
         }
     }
 
+    /// The code of this asset as a result.
+    pub fn asset_code(&self) -> Option<String> {
```
kbacha: Interesting. This would have been useful for the trades I believe as well. I wonder if there's another name that is more indicative that this will be optional. Trying to understand `code` for `asset_code`

resources/src/payment_path.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Source asset specified in the search that found this path
+    pub fn source_asset<'a>(&'a self) -> &'a AssetIdentifier {
+        &self.source_asset
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct IntermediatePaymentPath {
+    path: Vec<AssetIdentifier>,
```
kbacha: Dang, this works out so well.

PR #39, opened 2018-03-12
Endpoint trait should map to http request struct.

### LINKED ISSUE #37: Account Endpoint -> http::Request

kbacha:
This adds the request body as an associated type to the endpoint. In
some cases this will be a unit (as with the account details that lacks a
body). Other than that, this also maps some of the error types from http
into our own.

resolves #37

### DIFF
--- a/client/src/endpoint/mod.rs
+++ b/client/src/endpoint/mod.rs
@@ -1,14 +1,15 @@
 //! This module contains the various end point definitions for stellar's horizon
 //! API server.
 use error::Result;
-use hyper::Uri;
 use serde::Deserialize;
+use http;
 
 mod account;
 pub use self::account::AccountDetails;
 
 pub(crate) trait EndPoint<'de> {
     type Response: Deserialize<'de>;
+    type RequestBody;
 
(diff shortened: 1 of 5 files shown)


client/src/endpoint/mod.rs
```diff
@@ -1,14 +1,15 @@
...
-use hyper::Uri;
 use serde::Deserialize;
+use http;
 
 mod account;
 pub use self::account::AccountDetails;
 
 pub(crate) trait EndPoint<'de> {
     type Response: Deserialize<'de>;
+    type RequestBody;
```
correlator: Want to make sure I'm understanding this correctly.  An Endpoint must have a response that implements deserialize.  It must also have a request body that can be anything.

correlator: I like this a lot.

kbacha: Yup, exactly.

We may want to put a limit on the request body but there is no limit on it in the http crate at the moment. I'm not sure how we'd capture a `Serialize` trait on it while also allowing optional. Maybe there's a way :)

PR #40, opened 2018-03-13
Implement Sync Client

kbacha:
This commit implements the synchronous client library utilizing the
reqwest http client. It maps the end point into http and then converts
it to a reqwest request. This is then issued to the server and parsed
into the resulting response type correctly.

I've also divided up the client library into two components, the sync
and async module to allow easy choice between them. Generally the sync
will be useful for the CLI but async may be better for the actual web
server itself.

Lastly, I have written examples issuing a get query to the account
details end point :)

### DIFF
--- a/client/src/client/async.rs
+++ b/client/src/client/async.rs
@@ -1,0 +1,169 @@
+//! This module contains the client for asynchronous communcation.
+
+use hyper;
+use hyper_tls::HttpsConnector;
+use http;
+use tokio_core::reactor::Handle;
+use error::{Error, Result};
+use super::{Host, HORIZON_TEST_URI, HORIZON_URI};
+
+/// A client that can issue requests to a horizon api.
+#[derive(Debug, Clone)]
+pub struct Client {
+    inner: hyper::Client<HttpsConnector<hyper::client::HttpConnector>>,
+    host: Host,
+}
+
+impl Client {
+    /// Constructs a new stellar client.
+    ///
+    /// ## Examples
+    ///
+    /// ```
+    /// # extern crate tokio_core;
+    /// # extern crate stellar_client;
+    /// # fn main() {
+    /// use tokio_core::reactor::Core;
+    /// use stellar_client::async::Client;
+    /// let core = Core::new().unwrap();
+    /// let client = Client::new("https://horizon-testnet.stellar.org", &core.handle()).unwrap();
+    /// # }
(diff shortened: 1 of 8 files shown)


client/src/client/sync.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// ## Examples
+    ///
+    /// ```
+    /// use stellar_client::sync::Client;
+    /// use stellar_client::endpoint::AccountDetails;
+    /// let client = Client::horizon_test().unwrap();
+    /// let endpoint =
+    ///     AccountDetails::new("GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ");
+    /// let account = client.request(endpoint).unwrap();
+    /// assert_eq!(account.id(), "GCEZWKCA5VLDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ");
```
correlator: Wait whaaaaa 😻

client/src/client/sync.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// ```
+    pub fn request<E>(&self, endpoint: E) -> Result<E::Response>
+    where
+        E: EndPoint,
+    {
+        let request = endpoint.into_request(&self.uri())?;
+        let request = Self::http_to_reqwest(request);
+        let response = self.inner.execute(request)?;
+        if response.status().is_success() {
+            let resp: E::Response = serde_json::from_reader(response)?;
```
correlator: How does this know to use the deserializer from the DeserializeOwned associated type?

client/src/client/async.rs
```diff
@@ -0,0 +1,169 @@
+//! This module contains the client for asynchronous communcation.
```
correlator: should we make a note somewhere that this is not active i.e. does not implement a request method?

client/src/client/sync.rs
kbacha: Because the associated type `E::Response` must implement the trait `DeserializeOwned` so it calls the `deserialize` method that's on the `Account` struct.

client/src/client/async.rs
kbacha: 🤔 
not yet, we haven't published the crate yet and I plan on trying again soon.

PR #41, opened 2018-03-14
Scott/operations

correlator:
I have gone at this about 5 different ways and I think I'm happy with this data structure and organization.  Figured it would be good to check in on the pattern for a single operation type before going too deep.


resources/src/operation/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+
+impl<'de> Deserialize<'de> for Operation {
+    fn deserialize<D>(d: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let rep: IntermediateOperation = IntermediateOperation::deserialize(d)?;
+        match &rep.operation_type[..] {
+            "create_account" => {
+                if rep.account.is_none() || rep.funder.is_none() || rep.starting_balance.is_none() {
```
correlator: I think this logic will go in a `new` function on each OperationDetail

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    id: i64,
+    paging_token: String,
+    type_i: u32,
+    operation_detail: OperationDetail,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub enum OperationDetail {
+    /// Create Account operation represents a new account creation.
+    CreateAccount(CreateAccountFields),
```
kbacha: I don't like the `Fields` suffix on these. Wouldn't just `CreateAccount` be fine? Or possibly `CreateAccountOp`? It just feels like there's a better descriptor although I'm not sure what it might be.

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+/// Operations are objects that represent a desired change to the ledger: payments, offers to
+/// exchange currency, changes made to account options, etc. Operations are submitted to the
+/// Stellar network grouped in a Transaction.
+
+#[derive(Debug, Serialize)]
+pub struct Operation {
+    id: i64,
+    paging_token: String,
+    type_i: u32,
+    operation_detail: OperationDetail,
```
kbacha: aren't they called types in the stellar documentation? I'd rather use that nomenclature here.

If we do use `detail` let's just call it `detail` instead of prepending `operation_` to it. Makes it a little shorter.

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+            OperationDetail::CreateAccount(ref create_account_fields) => {
+                Some(create_account_fields.funder())
+            }
+            _ => None,
+        }
+    }
+
+    /// Returns the starting_balance of the account in the operation.
+    /// Returns None if the operation does not have a starting_balance attribute.
+    pub fn starting_balance(&self) -> Option<Amount> {
```
kbacha: These methods feel like they may pollute the operation struct. It might be better to just let clients match on the operation type and take it out to work with it there:

```ruby
if let OperationType::CreateAccount(op) = operation.op_type() {
    // Do something with op
    op.starting_balance()
}
```

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /// Specifies the type of operation, See “Types” section below for reference.
+    pub fn type_i(&self) -> u32 {
+        self.type_i
+    }
+
+    /// Returns true if the operation is a create_account operation
+    pub fn is_create_account(&self) -> bool {
+        match self.operation_detail {
```
kbacha: It's up to you but I sometimes like the `if let` syntax in these cases

```rust
if let OperationDetail::CreateAccount(_) = self.operation_detail() {
  true
} else {
  false
}
```

It's an extra line but I find the form a little more eloquent. And if we can shorten the words:

```rust
if let Detail::CreateAccount(_) = self.detail() {
  true
} else {
  false
}
```

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+
+/// Operations are objects that represent a desired change to the ledger: payments, offers to
+/// exchange currency, changes made to account options, etc. Operations are submitted to the
+/// Stellar network grouped in a Transaction.
+
+#[derive(Debug, Serialize)]
+pub struct Operation {
+    id: i64,
+    paging_token: String,
+    type_i: u32,
```
kbacha: We should be able to map an actual enum to this instead of just the same integer? Or just omit it and serialize to it via a match against the type.

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    operation_detail: OperationDetail,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub enum OperationDetail {
```
kbacha: Since `Type` might be reserved, we can probably just call this `Detail` to shorten it. Or perhaps `OpDetail`?

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+    /// Create Account operation represents a new account creation.
+    CreateAccount(CreateAccountFields),
+    /// A payment operation represents a payment from one account to another. This payment can be
+    /// either a simple native asset payment or a fiat asset payment.
+    Payment(PaymentFields),
+}
+
+#[derive(Debug, Deserialize)]
+pub struct IntermediateOperation {
```
kbacha: I'd move this closer to the deserialize impl

resources/src/operation/mod.rs
kbacha: Oh, a good word might be `Action` since that's what they are? Or `OpType` just to keep it shorter.

resources/src/operation/create_account.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+impl CreateAccountFields {
+    pub fn new(
+        account: Option<String>,
+        funder: Option<String>,
+        starting_balance: Option<Amount>,
+    ) -> Result<CreateAccountFields, String> {
+        if account.is_none() || funder.is_none() || starting_balance.is_none() {
+            Err("Account, funder and starting balance required.".to_string())
```
kbacha: This error doesn't look tested.

Also, if they can never be none, why even allow it in the type? If we make `new` a public function, we probably don't want to insert required optionals.

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+#[derive(Debug, Deserialize, Serialize)]
+struct SetOptions {
+    signer_key: String,
+    signer_weight: u8,
+    master_key_weight: u8,
+    low_threshold: u32,
+    med_threshold: u32,
+    high_threshold: u32,
+    home_domain: String,
+    set_flags: Option<Vec<u32>>,
```
kbacha: Does this need both sets of flag arrays for the internal type? Also, can we map the flags to an enum? Possibly even do the same with account if we haven't already?

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+/// Removes the account and transfers all remaining XLM to the destination account.
+#[derive(Debug, Deserialize, Serialize)]
+struct AccountMerge {
+    account: String,
+    into: String,
+}
+
+/// Runs inflation
+#[derive(Debug, Deserialize, Serialize)]
+struct Inflation {}
```
kbacha: Can't the `Inflation` enum value simply not have an inner value?

resources/src/operation/create_account.rs
correlator: They are options because at the moment of deserialization we don't know what type they will be.  That means all the fields in intermediate_operation must be options.  Those options then must be validated and unwrapped for each error type.  It feels better organized to put this logic in a new method separated into each OpDetail's mod rather than making the Operation deserializer handle all that logic.

resources/src/operation/mod.rs
correlator: You are correct.

resources/src/operation/create_account.rs
kbacha: I think that `new` should constitute creating a new instance of the struct. Passing deserialization to the new command as a `pub` function is creating complexity that doesn't need to be there. It should be captured in the deserialization process itself.

Issue #42, opened 2018-03-15
Implement cursor parsing for embedded resources


kbacha:
As part of the HAL specification for the response formats, the cursor is specified inside a URL in the `_links` object. We need to parse this out and provide the cursor so that we can effectively call `next` or `prev` on an end point.

PR #43, opened 2018-03-15
Implement the all assets end point

kbacha:
This implements the all assets endpoint which can be found documented
here:
https://www.stellar.org/developers/horizon/reference/endpoints/assets-all.html

In order to make this one work a deserializable `Records<T>` struct
needed to be constructed. This generally works but is lacking support
for parsing out the cursors. This has created the issue #42.

### DIFF
--- a/client/src/endpoint/asset.rs
+++ b/client/src/endpoint/asset.rs
@@ -1,0 +1,189 @@
+use error::Result;
+use std::str::FromStr;
+use stellar_resources::Asset;
+use super::{EndPoint, Order, Records};
+use http::{Request, Uri};
+
+/// Represents the all assets end point for the stellar horizon server. The endpoint
+/// will return all assets filtered by a myriad of different query params.
+///
+/// https://www.stellar.org/developers/horizon/reference/endpoints/assets-all.html
+///
+/// ## Examples
+///
+/// #### Asset code
+///
+/// Fetches all records for a given asset code.
+///
+/// ```
+/// # use stellar_client::sync::Client;
+/// # use stellar_client::endpoint::AllAssets;
+/// #
+/// let client      = Client::horizon_test().unwrap();
+/// let endpoint    = AllAssets::default().asset_code("USD");
+/// let records     = client.request(endpoint).unwrap();
+/// #
+/// # assert!(records.records().len() > 0);
+/// # assert_eq!(records.records()[0].code(), "USD");
+/// ```
+///
+/// #### Issuer
+///
+/// Fetches all records for a given asset issuer.
+///
+/// ```
+/// # use stellar_client::sync::Client;
+/// # use stellar_client::endpoint::AllAssets;
+/// #
+/// let client = Client::horizon_test().unwrap();
+/// # let endpoint = AllAssets::default().asset_code("USD");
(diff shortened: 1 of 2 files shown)


client/src/endpoint/asset.rs
```diff
@@ -0,0 +1,190 @@
...
+/// ```
+/// # use stellar_client::sync::Client;
+/// # use stellar_client::endpoint::AllAssets;
+/// #
+/// let client      = Client::horizon_test().unwrap();
+/// let endpoint    = AllAssets::default().asset_code("USD");
+/// let records     = client.request(endpoint).unwrap();
+/// #
+/// # assert!(records.records().len() > 0);
+/// # assert_eq!(records.records()[0].code(), "USD");
```
correlator: 💪

client/src/endpoint/asset.rs
```diff
@@ -0,0 +1,190 @@
...
+///
+/// Fetches all records for a given asset issuer.
+///
+/// ```
+/// # use stellar_client::sync::Client;
+/// # use stellar_client::endpoint::AllAssets;
+/// #
+/// let client = Client::horizon_test().unwrap();
+/// let endpoint = AllAssets::default()
+///     .asset_issuer("GA2LJYWSEZWZYIPMM4G6JXHM3PRHIEAD2TIR4IVJFZIAECYO7RDQJWOM");
```
correlator: They occasionally reset the testnet https://www.stellar.org/developers/guides/concepts/test-net.html

client/src/endpoint/asset.rs
```diff
@@ -0,0 +1,190 @@
...
+            .asset_code("CODE")
+            .asset_issuer("ISSUER")
+            .cursor("CURSOR")
+            .limit(123)
+            .order(Order::Desc);
+        let req = ep.into_request("https://www.google.com").unwrap();
+        assert_eq!(req.uri().path(), "/assets");
+        assert_eq!(
+            req.uri().query(),
+            Some("asset_code=CODE&asset_issuer=ISSUER&order=desc&cursor=CURSOR&limit=123")
```
correlator: 🕺 💃

client/src/endpoint/mod.rs
```diff
@@ -17,3 +20,90 @@ pub trait EndPoint {
...
+    pub(crate) fn to_param(&self) -> String {
+        match *self {
+            Order::Asc => "asc".to_string(),
+            Order::Desc => "desc".to_string(),
+        }
+    }
+}
+
+/// A struct that represents a set of records that comes from an embedded resource.
+#[derive(Debug)]
```
correlator: Well done.  Maybe a short comment about how resources from the stellar API are nested under _embedded and records. 🤷‍♂️

client/src/endpoint/asset.rs
kbacha: grrr

kbacha: It is an account from the examples but yea, i don't want a flaky test.

client/src/endpoint/mod.rs
kbacha: I might add that comment to the embedded struct/records but not to the records abstraction.

PR #44, opened 2018-03-16
Scott/operations

### LINKED ISSUE #8: add operation resource
The operations type is central to the entirety of stellar. There are multiple "types" within the operation, each pertaining to different things that need to be done.
https://www.stellar.org/developers/horizon/reference/resources/operation.html

There is likely going to need to be a builder that can assemble them into a transaction.

These are, of course, interesting, because they are likely XDRs and not JSON.

correlator:
This one got a little out of hand and pry should have been several smaller PRs.  Anyhow, this contains structs and deserializers for all operation types and resolves #8

### DIFF
--- a/resources/src/operation/manage_data.rs
+++ b/resources/src/operation/manage_data.rs
@@ -1,0 +1,48 @@
+/// Set, modify or delete a Data Entry (name/value pair) for an account.
+
+#[derive(Debug, Deserialize)]
+pub struct ManageData {
+    name: String,
+    value: String,
+}
+
+impl ManageData {
+    /// Creates a new ManageData
+    pub fn new(name: String, value: String) -> ManageData {
+        ManageData {
+            name: name,
+            value: value,
+        }
+    }
+
+    /// The key of the data value to update
+    pub fn name(&self) -> &String {
+        &self.name
+    }
+
+    /// The new data value associated with the named key
+    pub fn value(&self) -> &String {
+        &self.value
+    }
+}
+
+#[cfg(test)]
+mod manage_data_tests {
+    use serde_json;
+    use operation::{Operation, OperationDetail};
+
+    fn manage_data_json() -> &'static str {
+        include_str!("../../fixtures/operations/manage_data.json")
+    }
+
+    #[test]
+    fn it_parses_manage_data_from_json() {
+        let operation: Operation = serde_json::from_str(&manage_data_json()).unwrap();
+        assert!(operation.is_manage_data());
+        assert_eq!(operation.type_i(), 10);
+        if let &OperationDetail::ManageData(ref account_details) = operation.detail() {
+            assert_eq!(account_details.name(), "lang");
+            assert_eq!(account_details.value(), "aW5kb25lc2lhbg==");
+        }
+    }
+}
--- a/resources/src/operation/manage_offer.rs
+++ b/resources/src/operation/manage_offer.rs
@@ -1,0 +1,91 @@
+use amount::Amount;
+use asset::AssetIdentifier;
+use offer::PriceRatio;
+
+/// A “Manage Offer” operation can create, update or delete an offer to trade assets in the Stellar
+/// network. It specifies an issuer, a price and amount of a given asset to buy or sell.
+#[derive(Debug, Deserialize)]
+pub struct ManageOffer {
+    offer_id: i64,
+    selling: AssetIdentifier,
+    buying: AssetIdentifier,
+    amount: Amount,
+    #[serde(rename = "price_r")] price_ratio: PriceRatio,
+    price: Amount,
+}
+
+impl ManageOffer {
+    /// Creates a new ManageOffer
+    pub fn new(
+        offer_id: i64,
+        selling: AssetIdentifier,
+        buying: AssetIdentifier,
+        amount: Amount,
+        price_ratio: PriceRatio,
+        price: Amount,
+    ) -> ManageOffer {
+        ManageOffer {
+            offer_id: offer_id,
+            selling: selling,
+            buying: buying,
+            amount: amount,
+            price_ratio: price_ratio,
+            price: price,
+        }
+    }
+    /// Offer ID.
+    pub fn offer_id(&self) -> i64 {
+        self.offer_id
+    }
+
+    /// The identifier of the asset to sell.
+    pub fn selling(&self) -> &AssetIdentifier {
+        &self.selling
+    }
+
+    /// The identifier of the asset to buy.
+    pub fn buying(&self) -> &AssetIdentifier {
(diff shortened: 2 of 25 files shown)


resources/src/operation/mod.rs
```diff
@@ -0,0 +1,513 @@
...
+                    detail: OperationDetail::CreateAccount(
+                        self::create_account::CreateAccount::new(
+                            rep.account.unwrap(),
+                            rep.funder.unwrap(),
+                            rep.starting_balance.unwrap(),
+                        ),
+                    ),
+                })
+            }
+            "payment" => {
```
kbacha: You can probably match on the struct itself and simplify this code so that you don't need to do a lot of `is_none` checks. Since this will destructure into the variables on the left side of the match, you won't need to `unwrap` anything.

```rust
match *rep {
    IntermediateOperation {
        operation_type: "payment",
        from: Some(from),
        to: Some(to),
        asset_type: Some(asset_type),
        amount: Some(amount),
        ..
    } => {
        /* Create a new one since you know the values are good */
    },
    // If this matches you know its bad
    IntermediateOperation { operation_type: "payment", .. } => Err(...)
}
```

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,513 @@
...
+mod change_trust;
+mod create_account;
+mod create_passive_offer;
+mod inflation;
+mod manage_data;
+mod manage_offer;
+mod payment;
+mod path_payment;
+mod set_options;
+use self::account_merge::AccountMerge;
```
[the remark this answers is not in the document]
correlator: Good point, I will pub it up!

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,513 @@
...
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+pub struct IntermediateOperation {
+    id: i64,
+    paging_token: String,
+    #[serde(rename = "type")] operation_type: String,
+    type_i: u32,
```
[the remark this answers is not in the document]
correlator: It's true that this is redundant, but I would like to keep the API represented.  Mabe can tuck it into the subtypes.

resources/src/operation/mod.rs
correlator: Some of the operations have the exact same signature including ManageOffer and CreatePassiveOffer, so, unfortunately, we will not be able to deterministically switch on operation signature.

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,513 @@
...
+                if rep.offer_id.is_none() || rep.buying_asset_type.is_none()
+                    || rep.selling_asset_type.is_none() || rep.amount.is_none()
+                    || rep.price_ratio.is_none() || rep.price.is_none()
+                {
+                    Err(de::Error::custom(
+                        "Missing required params for create passive offer operation",
+                    ))?;
+                }
+                let buying_asset_identifier = AssetIdentifier::new(
+                    &rep.buying_asset_type.unwrap(),
```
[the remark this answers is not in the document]
correlator: there would be a lot of those in this file.

resources/src/operation/manage_data.rs
```diff
@@ -0,0 +1,48 @@
+/// Set, modify or delete a Data Entry (name/value pair) for an account.
+
+#[derive(Debug, Deserialize)]
```
kbacha: Do each of the operation details structs need this derivation?

resources/src/operation/manage_data.rs
```diff
@@ -0,0 +1,48 @@
+/// Set, modify or delete a Data Entry (name/value pair) for an account.
+
```
kbacha: Extra new line

resources/src/operation/manage_offer.rs
```diff
@@ -0,0 +1,91 @@
...
+    use operation::{Operation, OperationDetail};
+    use super::*;
+
+    fn manage_offer_json() -> &'static str {
+        include_str!("../../fixtures/operations/manage_offer.json")
+    }
+
+    #[test]
+    fn it_parses_a_manage_offer_from_json() {
+        let operation: Operation = serde_json::from_str(&manage_offer_json()).unwrap();
```
kbacha: These tests are actually of the operation details and not of any of the code inside here. Perhaps they should be moved to the operation file.

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+        let rep: IntermediateOperation = IntermediateOperation::deserialize(d)?;
+
+        let operation_detail: Option<OperationDetail> = match rep {
+            IntermediateOperation {
+                account: Some(account),
+                funder: Some(funder),
+                starting_balance: Some(starting_balance),
+                ..
+            } => {
+                if &rep.operation_type[..] == "create_account" {
```
kbacha: You shouldn't need the `[..]`

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+            }
+        };
+
+        match operation_detail {
+            Some(operation_detail) => Ok(Operation {
+                id: rep.id,
+                paging_token: rep.paging_token,
+                detail: operation_detail,
+            }),
+            None => Err(de::Error::custom("Invalid operation type."))?,
```
kbacha: Perhaps we should include the specified type here inside the string?

kbacha: I also don't think you need the `?` here as this is the return value.

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+            } => {
+                if &rep.operation_type[..] == "manage_data" {
+                    Some(OperationDetail::ManageData(
+                        self::manage_data::ManageData::new(name, value),
+                    ))
+                } else {
+                    None
+                }
+            }
+            _ => {
```
kbacha:
```rust
IntermediaOperation {
    operation_type: ref op_type,
} if op_type == "inflation" } => Some(OperationDetail::Inflation),
```

resources/src/operation/mod.rs
```diff
@@ -0,0 +1,520 @@
...
+                source_amount: Some(source_amount),
+                source_max: Some(source_max),
+                ..
+            } => {
+                if &rep.operation_type[..] == "path_payment" {
+                    let destination_asset_identifier = AssetIdentifier::new(
+                        &asset_type,
+                        rep.asset_code,
+                        rep.asset_issuer,
+                    ).map_err(|_| {
```
kbacha: This error comes back with a string. You should just map it into the custom error instead of assuming what the error is.

PR #45, opened 2018-03-17
Implement cursor links for records

kbacha:
When multiple records are returned there are links embedded in them.
These links provide pagination cursors to the next and previous page.
We now parse these values out of the uris and provide them in order to
page to the next end point.

### DIFF
--- a/client/src/endpoint/records.rs
+++ b/client/src/endpoint/records.rs
@@ -1,0 +1,140 @@
+use http;
+use serde::de::{Deserialize, DeserializeOwned, Deserializer};
+use std;
+
+/// A struct that represents a set of records returned from the horizon api.
+///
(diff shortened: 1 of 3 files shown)


client/src/endpoint/records.rs
```diff
@@ -0,0 +1,140 @@
...
+    href: String,
+}
+
+impl Href {
+    fn cursor(&self) -> String {
+        // Any error should just result in an empty string cursor
+        if let Ok(uri) = self.href.parse::<http::Uri>() {
+            if let Some(queries) = uri.query() {
+                if let Some(query) = queries.split("&").find(|q| q.starts_with("cursor=")) {
+                    return query.replacen("cursor=", "", 1);
```
correlator: very clever solution.

PR #46, opened 2018-03-18
Operations refactor

kbacha:
### First the tests
The tests that were in each file for the different operations were
actually testing the details parsing and not the structs explicitly.
Since the tests mostly involve the entire operation module, I'm moved
them into their own file `test.rs` that is only built in test mode.

### Next, the match
To simplify how the matching works, I've refactored it into a single
match that errors on types that don't match their type. The key to
making this match work was changing the intermediate type field from a
String to a &'a str. This allows rust to perform a literal match as
opposed to a conditional one.

Another aspect of the change was to return early with an Err if a match
determines the error. This allows us to return an explicit operation
detail from the match and return from the function if it errors.

This isn't explicitly shorter but it is one nesting level shallower. 

### Lastly, some renames
Some of the names are very long and redundant with the module that they
are contained in. By renaming them privately, the file itself can be a
bit less redundant while not changing the public interface at all. It's
pretty easy to assume that the `Intermediate` struct is representative
of the operation as it used in the operation module.

Similarly, `OperationDetail` is renamed to `OperationKind` and is aliased
to `Kind` privately.

I've also renamed `operation_type` to `kind` as it means the
same thing with many fewer characters.


resources/src/operation/mod.rs
```diff
@@ -252,95 +255,112 @@ impl<'de> Deserialize<'de> for Operation {
...
+                    to,
+                    destination_asset_identifier,
+                    amount,
+                    source_asset_identifier,
+                    source_amount,
+                    source_max,
+                ))
+            }
+            IntermediateOperation {
+                operation_type: "path_payment",
```
kbacha: I'm debating moving all the error cases to the end of the match block. What do you think?

resources/src/operation/mod.rs
```diff
@@ -398,132 +429,140 @@ impl<'de> Deserialize<'de> for Operation {
...
-                    None
-                }
-            }
-            _ => {
-                if &rep.operation_type[..] == "inflation" {
-                    Some(OperationDetail::Inflation)
-                } else {
-                    None
-                }
+                return Err(de::Error::custom(
```
kbacha: I'm debating moving all the errors to the bottom of the match. Thoughts?

resources/src/asset.rs
```diff
@@ -246,9 +246,8 @@ impl<'de> Deserialize<'de> for Asset {
     {
         let rep: IntermediateAsset = IntermediateAsset::deserialize(d)?;
         let asset_identifier: Result<AssetIdentifier, D::Error> =
-            AssetIdentifier::new(&rep.asset_type, rep.asset_code, rep.asset_issuer).map_err(|_| {
-                de::Error::custom("Code and issuer are required for non-native assets")
-            });
+            AssetIdentifier::new(&rep.asset_type, rep.asset_code, rep.asset_issuer)
+                .map_err(|err| de::Error::custom(err));
```
correlator: 👌

resources/src/operation/mod.rs
```diff
@@ -252,95 +260,107 @@ impl<'de> Deserialize<'de> for Operation {
...
+                    destination_asset_identifier,
+                    amount,
+                    source_asset_identifier,
+                    source_amount,
+                    source_max,
+                ))
             }
-            IntermediateOperation {
-                operation_type,
+            Intermediate {
```
correlator: This is so great

resources/src/operation/mod.rs
correlator: I like how it is now because it offers additional information about the kind it was expecting.  If we move it down here, maybe we should include the data from Intermediate in the message.

kbacha: Sorry, I meant moving the clauses down, not removing the extra error messaging.

PR #49, opened 2018-03-21
Implement a specific body type

kbacha:
So that we can be a little more specific in how an endpoint works, the
trait for an endpoint must now produce a request with a body of type
`Body`. This is an enum that can capture the idea of `None` but also
potentially capture the idea that there is a body in the future.
Currently, the only end point are `GET` endpoints so the body doesn't
matter but this removes the arbitrary generic typing we had around it.


correlator:
I dig it.  You're looking around the corner here!

PR #66, opened 2018-03-24
Add all effect resources.

### LINKED ISSUE #13: Add effects
A successful operation may return effects

https://www.stellar.org/developers/horizon/reference/resources/effect.html

correlator:
An effect is a change in the ledger that is created by an
operation.  We collect these resources to make our application
eventually consistent with the state of the world in the stellar ledger.

This resource is modeled in the same way that operations are modeled
creating for consistency.  This includes shortening
"EffectDetail" to "Kind", the simplified matching patterns, and the same
error handling patterns.

Resolves #13

### DIFF
--- a/resources/src/effect/account_thresholds_updated.rs
+++ b/resources/src/effect/account_thresholds_updated.rs
@@ -1,0 +1,41 @@
+/// This effect can be the result of a set options operation and represents
+/// the fact that an account's weight thresholds have changed.
+#[derive(Debug, Deserialize)]
+pub struct AccountThresholdsUpdated {
+    account: String,
+    low: u32,
+    med: u32,
+    high: u32,
+}
+
+impl AccountThresholdsUpdated {
+    /// Creates a new AccountThresholdsUpdated
+    pub fn new(account: String, low: u32, med: u32, high: u32) -> AccountThresholdsUpdated {
+        AccountThresholdsUpdated {
+            account: account,
+            low: low,
+            med: med,
+            high: high,
+        }
+    }
+
+    /// The public address of the account that had its thresholds updated.
+    pub fn account(&self) -> &String {
+        &self.account
+    }
+
+    /// The sum weight for the low threshold.
+    pub fn low(&self) -> u32 {
+        self.low
+    }
+
+    /// The sum weight for the medium threshold.
+    pub fn med(&self) -> u32 {
+        self.med
+    }
+
+    /// The sum weight for the high threshold.
+    pub fn high(&self) -> u32 {
+        self.high
+    }
+}
--- a/resources/src/effect/mod.rs
+++ b/resources/src/effect/mod.rs
@@ -1,0 +1,562 @@
+use amount::Amount;
+use asset::{AssetIdentifier, Flag};
+use serde::{de, Deserialize, Deserializer};
+mod account_created;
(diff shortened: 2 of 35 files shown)


correlator:
I clicked update branch here and it added that merge commit.  I'll squash that away before rebase/merging this branch

resources/src/effect/account_thresholds_updated.rs
```diff
@@ -0,0 +1,46 @@
+/// This effect can be the result of a set options operation and represents
+/// the fact that an account's weight thresholds have changed.
+#[derive(Debug, Deserialize)]
+pub struct AccountThresholdsUpdated {
+    account: String,
+    low_threshold: u32,
```
kbacha: Could we possibly rename these to just `low` `medium` and `high` since threadshold is in the struct name?

resources/src/effect/account_thresholds_updated.rs
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+
+    /// The public address of the account that had its thresholds updated.
+    pub fn account(&self) -> &String {
+        &self.account
+    }
+
+    /// The sum weight for the low threshold.
+    pub fn low_threshold(&self) -> u32 {
```
kbacha: We could also just rename the accessors if we don't rename the attributes.

resources/src/effect/mod.rs
```diff
@@ -0,0 +1,562 @@
+use amount::Amount;
+use asset::{AssetIdentifier, Flag};
+use serde::{de, Deserialize, Deserializer};
+mod account_created;
```
kbacha: These modules make me think that it might be cleaner to just have a module per "type" of effect. This would mean `account`, `signer`, `trade`, `trustline`. Then have a handful of structs in each of those files.

resources/src/effect/mod.rs
```diff
@@ -0,0 +1,562 @@
...
+    paging_token: String,
+    kind: Kind,
+}
+
+/// Each effect type is representing by a kind and captures data specific to that
+/// type within it's newtype.
+#[derive(Debug, Deserialize)]
+pub enum EffectKind {
+    /// An effect representing the fact that an account was created
+    AccountCreated(AccountCreated),
```
kbacha: Going along with the idea of modules per type of effect, you might have an enum for that as well. It becomes more nested and that might be bad as well, but I could see something like:

```rust
enum Kind {
  Account(account::Kind)
  ..
}
```

and then in the account module:
```rust
enum Kind {
  Created(AccountCreated),
  Removed(AccountRemoved),
}
```

Not sure that helps, exactly, but it does allow for potentially easy checking what the affect applies to:

```rust
match effect.kind() {
  Kind::Account(_) => {},
  Kind::Signer(_) => ...
}
```

resources/src/effect/mod.rs
correlator: I think that you are correct, and I also think that we can do something similar with operations.  Given that this is such a huge chunk of code I would prefer to get this in and do those refactors in a separate PR.  What are your thoughts?

kbacha: Sure, we can do it as a separate PR. Can you make an issue for it for tracking?

PR #69, opened 2018-03-25
Refactor account endpoint into module

### LINKED ISSUE #63: Refactor account endpoint into a module
There are several new endpoints that need to be created. It would make sense to namespace these into a module so that we can keep them together and not need obtuse struct names for every endpoint.

So, instead of
```rust
use endpoint::AllAccounts
```
as an endpoint we'll do
```rust
use endpoint::account::All
```

This can then be mirrored for others as well:
```rust
use endpoint::ledger::All
use endpoint::ledger::Operations
```

kbacha:
I felt capturing the endpoints related to accounts and assets made sense
in a module. These modules are where related details, all, and "for x"
endpoints will wind up. I think this may provide a slightly more
ergonomic method for using and separating the concerns and logically
grouping the related endpoints. It allows for endpoints named like:

```
asset::All;
account::Effects;
account::Details;
account::Operations
ledger::All;
ledger::Details;
ledger::Operations;
```

These can be aliased in specific usecases when AllAssets is nicer and
may conflict but keeps the code easier to read within their
implementations.

```rust
use stellar_client::endpoint::asset::All as AllAssets;
```

Or just the module can be required (as I've put in the doc-tests):

```rust
use stellar_client::endpoint::asset;

let endpoint = asset::All::default();
```

Second change that's pertinent, I moved the doctests for the asset
module to the individual builder methods. I felt this better documents
each of the setters, while leaving a single `default` implementation at
the heading of the struct.

image

Fixes #63

### DIFF
--- a/client/src/endpoint/mod.rs
+++ b/client/src/endpoint/mod.rs
@@ -4,12 +4,10 @@ use error::Result;
 use serde::de::DeserializeOwned;
 use http;
 
-mod account;
-mod asset;
+pub mod account;
+pub mod asset;
 mod records;
 pub use self::records::Records;
-pub use self::account::AccountDetails;
-pub use self::asset::AllAssets;
 
 /// Represents the body of a request to an EndPoint.
 #[derive(Debug)]
--- a/client/src/client/sync.rs
+++ b/client/src/client/sync.rs
@@ -104,10 +104,10 @@ impl Client {
     ///
     /// ```
     /// use stellar_client::sync::Client;
-    /// use stellar_client::endpoint::AccountDetails;
+    /// use stellar_client::endpoint::account;
     /// let client = Client::horizon_test().unwrap();
     /// let endpoint =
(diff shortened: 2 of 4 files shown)


client/src/endpoint/mod.rs
```diff
@@ -4,12 +4,10 @@ use error::Result;
 use serde::de::DeserializeOwned;
 use http;
 
-mod account;
-mod asset;
+pub mod account;
+pub mod asset;
```
correlator: Why don't we need any documentation here?  Looks like things are silenced here https://github.com/kbacha/stellar-sdk/blob/master/client/src/lib.rs#L1  any particular reason for this?

kbacha: It's not silenced. I moved the documentation to the header of the asset and account mods
https://github.com/kbacha/stellar-sdk/pull/69/files#diff-a7ef25ce71dce3bbe0f8bed2034b991cR1

PR #70, opened 2018-03-25
Initial implementation of the cli

kbacha:
There is still much to do but this outlines the initial format that I'm
thinking of for the CLI. Basically, you can use sub-commands to request
details about different aspects of the horizon API. I've only
implemented the account and there will need to be more working to make
it not quite as messy as we add more but this is the first step.

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,2 @@
 [workspace]
-members = ["resources", "client"]
+members = ["resources", "client", "cli"]
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "stellar-cli"
+version = "0.1.0"
+authors = ["Kevin Bacha <chewbacha@gmail.com>"]
+
+[[bin]]
+name = "stellar"
+path = "src/main.rs"
+
+[dependencies]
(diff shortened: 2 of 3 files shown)


robertDurst:
Looks great! Noob questions here : 

a) what is the use of and purpose of the workspace
b) is there a better way to do subcommands without the crazy nested indents?

kbacha:
Sure thing!

1. The workspace allows us to work from the root directory with with `cargo` commands but have smaller crates. In this case, we have a resources crate, a client crate, and a cli crate. We can keep the interfaces completely separate this way and document them independently. I'm still debating on the correct means to segregate all of it but it is a common pattern.

An additional benefit is that artifacts from one crate are compiled separately which means that recompiling should be significantly faster. In other words, if you only change the client you won't need to recompile the resources.

https://doc.rust-lang.org/book/second-edition/ch14-03-cargo-workspaces.html

2. I think that part of it I'd like to keep but I do want to move the handling of each command into functions to avoid the nesting. It's messy right now and I really do want to break it up and make it more legible. The match on the command though I actually like, i just want that to be the center of attention and not the other stuff. I plan on working on this a lot more, I just wanted to demonstrate what the command interface would be itself.

PR #67, opened 2018-03-24
Add all ledgers endpoint

robertDurst:
Modeled heavily off of All Assets, this endpoint represents the all ledgers endpoint for the stellar horizon server.

Since this is my first PR please critique mercilessly :)

Closes #57

PR #71, opened 2018-03-26
Fix the warnings from `cargo doc`

kbacha:
There are warnings in the rendering differences when running `cargo
doc`. This is because rust is switching from `hoedown` to `pulldown` as
a rendering client. The changes are minor but auto-links are no longer
correctly supported. If you want to put a link you need to wrap it with
`<` and `>` or put it in a standard markdown link: `[text](link)`.

Another minor difference that I added was putting an extra space above
raw links to separate them from the descriptive text a bit. I also
wrapped one of the links into text so that it reads more intuitively.

For testing this, you should be able to run `cargo doc` and see no more
warnings on this repo.

### DIFF
--- a/client/src/stellar_error.rs
+++ b/client/src/stellar_error.rs
@@ -3,10 +3,8 @@ use serde::{Deserialize, Deserializer};
 use std::fmt;
 
 /// A resource for the stellar horizon API specific error codes.
-/// These errors adhere to the Problem Details Standard
-/// https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-00
-/// and a list of possible erros can be found here
-/// https://www.stellar.org/developers/horizon/reference/errors.html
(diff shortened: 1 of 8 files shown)


robertDurst:
Awesome, I will have to fix the docs in #67

client/src/stellar_error.rs
```diff
@@ -3,10 +3,8 @@ use serde::{Deserialize, Deserializer};
 use std::fmt;
 
 /// A resource for the stellar horizon API specific error codes.
-/// These errors adhere to the Problem Details Standard
-/// https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-00
-/// and a list of possible erros can be found here
-/// https://www.stellar.org/developers/horizon/reference/errors.html
+/// These errors adhere to the [Problem Details Standard](https://tools.ietf.org/html/draft-ietf-appsawg-http-problem-00)
+/// and a list of possible erros can be found [on the Stellar website](https://www.stellar.org/developers/horizon/reference/errors.html)
```
correlator: nice

PR #79, opened 2018-03-27
Scott/create effects

### LINKED ISSUE #13: Add effects

correlator:
Adds effects and Resolves #13 

Effects are the byproduct of operations and represent changes in the ledger.  These changes typically involve some sort of CRUD like operation on a resource.  We therefore grouped effects into modules based on the resource that are updated.  This will allow users of the sdk to switch on effect kind and know what kind of resource will be updated by the change.


resources/src/effect/mod.rs
```diff
@@ -0,0 +1,564 @@
...
+#[derive(Debug, Deserialize)]
+pub enum EffectKind {
+    /// A collection of effects that represent updates to an account
+    Account(account::Kind),
+    /// A collection of effects that represent updates to an account signer
+    Signer(signer::Kind),
+    /// A collection of effects that represent updates to a trustline
+    Trustline(trustline::Kind),
+    /// An effect representing a trade being executed
+    Trade(Trade),
```
kbacha: Perhaps we should use a `Kind` here as well so that if an effect was added for trades that it would be extensible without breaking the public interface?

resources/src/effect/mod.rs
```diff
@@ -0,0 +1,564 @@
...
+
+    /// A paging token suitable for use as a cursor parameter.
+    pub fn paging_token(&self) -> &String {
+        &self.paging_token
+    }
+
+    /// Specifies the type of effect, See “Types” section below for reference.
+    pub fn type_i(&self) -> u32 {
+        match &self.kind {
+            &Kind::Account(ref account_kind) => match account_kind {
```
kbacha: You could capture the inner kind by ref and then deref it for the match. This removes all the little ampersands:

```rust

/// Specifies the type of effect, See “Types” section below for reference.
pub fn type_i(&self) -> u32 {
    match self.kind {
        Kind::Account(ref account_kind) => match *account_kind {
            account::Kind::Created(_) => 0,
            account::Kind::Removed(_) => 1,
            account::Kind::Credited(_) => 2,
            account::Kind::Debited(_) => 3,
            account::Kind::ThresholdsUpdated(_) => 4,
            account::Kind::HomeDomainUpdated(_) => 5,
            account::Kind::FlagsUpdated(_) => 6,
        },
        Kind::Signer(ref signer_kind) => match *signer_kind {
            signer::Kind::Created(_) => 10,
            signer::Kind::Removed(_) => 11,
            signer::Kind::Updated(_) => 12,
        },
        Kind::Trustline(ref trustline_kind) => match *trustline_kind {
            trustline::Kind::Created(_) => 20,
            trustline::Kind::Removed(_) => 21,
            trustline::Kind::Updated(_) => 22,
            trustline::Kind::Authorized(_) => 23,
            trustline::Kind::Deauthorized(_) => 24,
        },
        Kind::Trade(_) => 33,
    }
}
```

resources/fixtures/effects/account_created.json
```diff
@@ -0,0 +1,21 @@
...
+    }
+  },
+  "account": "GBS43BF24ENNS3KPACUZVKK2VYPOZVBQO2CISGZ777RYGOPYC2FT6S3K",
+  "id": "0000010157597659139-0000000002",
+  "paging_token": "141733924865-1",
+  "starting_balance": "10000000.0",
+  "type_i": 0,
+  "type": "account_created"
+}
+
```
kbacha: extraneous white space

PR #85, opened 2018-03-29
Ran cargo fmt

kbacha:
Since rust has released v1.25 the default cargo fmt settings have
changed (for the better). It looks like attribute annotations now end up
the second line. This resolves that.

### DIFF
--- a/client/src/endpoint/records.rs
+++ b/client/src/endpoint/records.rs
@@ -60,8 +60,10 @@ where
 /// https://www.stellar.org/developers/horizon/reference/responses.html
 #[derive(Deserialize)]
 struct Embedded<T> {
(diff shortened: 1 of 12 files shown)


robertDurst:
Interesting

client/src/endpoint/records.rs
```diff
@@ -60,8 +60,10 @@ where
 /// https://www.stellar.org/developers/horizon/reference/responses.html
 #[derive(Deserialize)]
 struct Embedded<T> {
-    #[serde(rename = "_embedded")] embedded: T,
-    #[serde(rename = "_links")] links: Links,
+    #[serde(rename = "_embedded")]
+    embedded: T,
```
correlator: 😻

correlator:
Love this update!

PR #94, opened 2018-03-30
Implement bad response parsing for client errors

### LINKED ISSUE #27: Implement error and error kind for resource parsing

kbacha:
If the request to stellar is incorrect we may get back a bad response.
This response has a defined type and we can parse it from that json type
into our known StellarErrors. Along the way I noticed that the "type"
field is actually a url and not a slug at all. I've corrected it.

I also renamed the `ErrorType` field into `Kind` and began using the
`FromStr` trait to parse the `Kind` instead of a custom method. It might
be useful to parse this into a separate field as well to return the URL
to visit in the future as reference.

fixes #27

### DIFF
--- a/client/fixtures/before_history_error.json
+++ b/client/fixtures/before_history_error.json
@@ -1,5 +1,5 @@
 {
-  "type": "before_history",
+  "type": "https://stellar.org/horizon-errors/before_history",
   "title": "Data Requested Is Before Recorded History",
   "status": 410,
   "detail": "This horizon instance is configured to only track a portion of the stellar network's latest history. This request is asking for results prior to the recorded history known to this horizon instance.",
--- a/client/src/client/sync.rs
+++ b/client/src/client/sync.rs
@@ -5,6 +5,7 @@ use http::{self, Uri};
 use error::{Error, Result};
 use endpoint::EndPoint;
 use super::{Host, HORIZON_TEST_URI, HORIZON_URI};
+use StellarError;
 use serde_json;
 
 /// A client that can issue requests to a horizon api in a synchronous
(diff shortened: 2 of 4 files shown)


client/fixtures/before_history_error.json
```diff
@@ -1,5 +1,5 @@
 {
-  "type": "before_history",
+  "type": "https://stellar.org/horizon-errors/before_history",
```
correlator: This is definitely more useful

client/src/client/sync.rs
```diff
@@ -185,4 +189,15 @@ mod tests {
...
         );
     }
+
+    #[test]
+    fn it_can_make_a_failed_request() {
+        use endpoint::account::Details;
+        let client = Client::horizon_test().unwrap();
+        let endpoint = Details::new("LDNRLN3RPRJMRZOX3Z6G5CHCGSNFHEYVXM3XOJMDS674JZ");
+        match client.request(endpoint).unwrap_err() {
+            Error::BadResponse(error) => assert_eq!(error.kind(), stellar_error::Kind::NotFound),
```
correlator: very nice.

client/src/stellar_error.rs
```diff
@@ -32,53 +32,43 @@ pub enum ErrorType {
...
-        Ok(StellarError {
-            error_type: type_enum,
-            title: title,
-            status: status,
-            detail: detail,
-            instance: instance,
+impl FromStr for Kind {
+    type Err = super::Error;
+
+    fn from_str(s: &str) -> Result<Self, super::Error> {
```
correlator: Is there an added benefit to keeping this outside of the deserializer?  I see you use this in a test, but could also use the same deserialize pattern we use in other resource files.  Curious to hear your thoughts.

kbacha: Well, it was captured outside the deserializer before. It was in the `error_from_str` method and we exposed a `new` function along with it for testing.

I haven't actually added to the public interface so much as moved it to a more specific Trait that's used in rust.

I do think it makes sense to `impl FromStr` when we can :)

PR #97, opened 2018-03-30
Capture the url provided from stellar

kbacha:
The stellar error response includes a useful URL that provides
additional information about what happened. This captures it and
includes it in the formatted display of the error.


client/src/stellar_error.rs
```diff
@@ -193,7 +228,7 @@ impl Error for StellarError {
 
 impl fmt::Display for StellarError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.write_str(self.description())
+        write!(f, "{}\n\nTo learn more: {}", self.detail, self.url)
```
correlator: 🌶

PR #99, opened 2018-03-30
Make the cursor doctest fallible

kbacha:
The current cursor tests were infallible since the limits between the
first page and second page were different. This meant that the next
cursor was always different. By using the same limit these tests can
actually fail if the cursor is not implemented.


correlator:
🌟

PR #100, opened 2018-03-30
[#91] Implements Transactions for account

### LINKED ISSUE #91: [endpoint] Transactions for account
This endpoint represents all transactions that affected a given account. This endpoint can also be used in streaming mode so it is possible to use it to listen for new transactions as that affect a given account as they get made in the Stellar network

```rust
use stellar_client::endpoint::account::Transactions;
```

https://www.stellar.org/developers/horizon/reference/endpoints/transactions-for-account.html

bringking:
Closes #91

### DIFF
--- a/client/src/endpoint/account.rs
+++ b/client/src/endpoint/account.rs
@@ -2,7 +2,8 @@
 use error::Result;
 use std::str::FromStr;
 use stellar_resources::Account;
-use super::{Body, EndPoint};
+use stellar_resources::Transaction;
+use super::{Body, EndPoint, Order, Records};
 use http::{Request, Uri};
 
 /// An endpoint that accesses a single accounts details.
@@ -30,7 +31,7 @@ impl EndPoint for Details {
 }
 
 #[cfg(test)]
-mod tests {
+mod details_tests {
     use super::*;
 
     #[test]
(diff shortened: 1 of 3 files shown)


client/src/endpoint/account.rs
```diff
@@ -43,3 +44,171 @@ mod tests {
...
+    /// let client   = Client::horizon_test().unwrap();
+    /// #
+    /// # // grab first page and extract cursor
+    /// # let transaction_ep   = transaction::All::default().limit(1);
+    /// # let txns             = client.request(transaction_ep).unwrap();
+    /// # let txn              = &txns.records()[0];
+    /// # let endpoint         = account::Transactions::new(txn.source_account());
+    /// # let first_page       = client.request(endpoint).unwrap();
+    /// # let cursor           = first_page.next_cursor();
+    /// # let account_id       = txn.source_account();
```
correlator: a little nit-picky, but can move this up a few rows to use when defining the endpoint.

client/src/endpoint/transaction.rs
```diff
@@ -4,6 +4,7 @@ use std::str::FromStr;
 use stellar_resources::Transaction;
 use super::{Body, EndPoint, Order, Records};
 use http::{Request, Uri};
+pub use super::account::Transactions as ForAccount;
```
correlator: NICE

bringking: That was a @kbacha special! He gave me the idea. Turns out nice, it matches the API in the JS sdk

PR #101, opened 2018-03-31
Additional work on the CLI

kbacha:
Adds a couple new sub commands related to transactions. Also splits out
the "action" of each command to functions and adds pagination by
listening to stdin. Still need to figure out what information is
important to display and how to render it effectively.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -2,17 +2,18 @@ extern crate clap;
 extern crate stellar_client;
 extern crate stellar_resources;
 
-use clap::{App, Arg, SubCommand};
+use clap::{App, AppSettings, Arg, SubCommand};
 use stellar_client::sync::Client;
-use stellar_client::endpoint::account;
 
-fn main() {
-    let matches = App::new("Stellar CLI")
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -21,32 +22,145 @@ fn main() {
...
+        match client.request(endpoint) {
+            Ok(records) => {
+                for txn in records.records().iter() {
+                    println!("ID:         {}", txn.id());
+                    println!("account id: {}", txn.source_account());
+                    println!("created at: {}", txn.created_at());
+                    println!("");
+                }
+                if next_page() {
+                    transactions(client, &matches, Some(records.next_cursor().to_string()))
```
correlator: 🕺

Interesting that there is always a next_cursor() value, but the API will send back a 400 if you call the `_links#next` url that horizon returns.  Nothing to do with your code, just writing down thoughts.

kbacha: Interesting, good point. It ends up just returning empty sets. I should break the loop when there's nothing.

PR #102, opened 2018-03-31
[#93] Transaction Details

### LINKED ISSUE #93: [endpoint] Transaction Details
The transaction details endpoint provides information on a single transaction. The transaction hash provided in the hash argument specifies which transaction to load.

```rust
use stellar_client::endpoint::transaction::Details;
```

https://www.stellar.org/developers/horizon/reference/endpoints/transactions-single.html

bringking:
Implements Transaction details. https://www.stellar.org/developers/horizon/reference/endpoints/transactions-single.html. Also includes an example and doc test.

Closes #93

### DIFF
--- a/client/src/endpoint/transaction.rs
+++ b/client/src/endpoint/transaction.rs
@@ -154,3 +154,63 @@ mod all_transactions_test {
         );
     }
 }
+
+/// Represents the details for a single transaction.
+///
+/// <https://www.stellar.org/developers/horizon/reference/endpoints/transactions-single.html>
+///
+/// ## Example
+///
+/// ```
+/// use stellar_client::sync::Client;
+/// use stellar_client::endpoint::transaction;
+///
+/// let client   = Client::horizon_test().unwrap();
+/// # let transaction_ep   = transaction::All::default().limit(1);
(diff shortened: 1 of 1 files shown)


client/src/endpoint/transaction.rs
```diff
@@ -154,3 +154,63 @@ mod all_transactions_test {
...
+///
+/// let client   = Client::horizon_test().unwrap();
+/// # let transaction_ep   = transaction::All::default().limit(1);
+/// # let txns             = client.request(transaction_ep).unwrap();
+/// # let txn              = &txns.records()[0];
+/// # let hash             = txn.hash();
+/// let endpoint = transaction::Details::new(hash);
+/// let txn      = client.request(endpoint).unwrap();
+/// #
+/// # assert_eq!(txn.hash(), hash);
```
correlator: Optional: could we use more expressive names here?  This isn't a huge deal and most of these lines are hidden in docs, but I think it would be easier to read with names like `transaction`, `transaction_hash`, `transaction_details` etc.

kbacha: I think the reason was that it interferes with the module that's being used.

PR #104, opened 2018-03-31
Implement Iterator for Records

kbacha:
So, the actual Records struct can't really be an iterator across all
records due to it's lack of knowledge about the client and the endpoint
it's part of. But, we can create a record iter for use with the sync client!

This commit implements that type. It requires some traits to be
implemented by the resource and the endpoint for it to work correctly.
It needs to be able to clone the endpoint and the resource. Also, the
endpoint must have an associated type of a `Record`. This allows us to
paginate based on the endpoint, cache the returned records, then iterate
through them.

This immediately allows us to paginate across the entire set of records
associated with whatever endpoint we want. This _should_ be lazy based
on the way that rust iterators work themselves as they are lazy by
default.

### DIFF
--- a/client/src/client/sync/iter.rs
+++ b/client/src/client/sync/iter.rs
@@ -1,0 +1,127 @@
+use endpoint::{Cursor, EndPoint, Records};
+use serde::de::DeserializeOwned;
+use super::Client;
+use error::Result;
+
+/// An iterator for records. Provides the ability to use the iterator
+/// in rust against records that are returned from the api.
+///
+/// # Examples
+///
+/// ```
+/// use stellar_client::{
+///     endpoint::asset,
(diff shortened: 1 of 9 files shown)


client/src/client/sync/iter.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+}
+
+impl<'a, T, E> Iterator for RecordIter<'a, T, E>
+where
+    E: EndPoint<Response = Records<T>> + Clone + Cursor<T>,
+    T: DeserializeOwned + Clone,
+{
+    type Item = Result<T>;
```
kbacha: I just realized that the error is never actually returned...

kbacha: fixed 😉

PR #105, opened 2018-04-01
Renames the trait EndPoint to IntoRequest

kbacha:
To follow along with what appear to be common naming conventions in rust
I'm proposing to rename the EndPoint trait to IntoRequest. This does not
change the module (I still like classifying it as endpoints at the
module lever) but I want to be explicit about what the trait does, not
what it's supposed to "be". Since the point of the trait is to consume
itself and produce a request, this trait name seems more appropriate.

see this issue in the rust-nursery:
https://github.com/rust-lang-nursery/api-guidelines/issues/28

warning: This is may be slightly controversial.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,6 +96,13 @@ nursery: https://github.com/rust-lang-nursery/rustfmt
 Naming and conventions to improve the public interface should be followed and considered. We want
 sdk to be as easy to use as possible.
 
+When developing a trait please consider that traits are more like verbs than nouns. As such the
+convention has been to utilize words like `Clone` instead of `Cloneable` and `Write` instead of
+`Writer`. Sometimes a noun might make the right word but generally we should favor describing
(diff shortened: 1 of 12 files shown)


correlator:
This could also warrant a mention in the style section of our contributing.md file.  Maybe that's overkill at this point, but worth thinking about.

CONTRIBUTING.md
```diff
@@ -96,6 +96,13 @@ nursery: https://github.com/rust-lang-nursery/rustfmt
 Naming and conventions to improve the public interface should be followed and considered. We want
 sdk to be as easy to use as possible.
 
+When developing a trait please consider that traits are more like verbs than nouns. As such the
+convention has been to utilize words like `Clone` instead of `Cloneable` and `Write` instead of
+`Writer`. Sometimes a noun might make the right word but generally we should favor describing
+what it does rather than what it is.
```
correlator: 👍

PR #106, opened 2018-04-02
Major enhancements to cli

kbacha:
Several things went into this change. Some was refactoring and some were
new enhancements.

*Added All Assets*

The first change was the most straight forward, the assets endpoint was
added. This end point features filtering via the command line args of
code and issuer.

*Switched to using the Iter*

The second change was the migration to using the iterator. This majorly
simplifies the code and eliminates the need for the recursion that was
there to begin with. I use the `enumerate` iterator to pair each line
with the index so that we can paginate.

*Functions return result types now*

Instead of duplicating the exception handling through out, instead we
just have each function return the client errors and handle it all at
the end.

*Choose the network!*

Another option that was added was the ability specify which network you
wished to join!

*Move the groups of functions to modules*

The file was feeling a little complex so I moved each group of commands
to their own module file. They were already modules so this was very
straightforward.

### DIFF
--- a/cli/src/account.rs
+++ b/cli/src/account.rs
@@ -1,0 +1,34 @@
+use clap::ArgMatches;
+use stellar_client::{sync, endpoint::{account, Order}, error::Result, sync::Client};
+use super::utils;
+
+pub fn details<'a>(client: Client, matches: &'a ArgMatches) -> Result<()> {
+    let id = matches.value_of("ID").expect("ID is required");
+    let endpoint = account::Details::new(id);
+    let account = client.request(endpoint)?;
+
+    println!("ID:       {}", account.id());
+    println!("Sequence: {}", account.sequence());
+
+    Ok(())
+}
+
+pub fn transactions<'a>(client: Client, matches: &'a ArgMatches) -> Result<()> {
+    let id = matches.value_of("ID").expect("ID is required");
+    let endpoint = account::Transactions::new(id).order(Order::Desc);
+    let iter = sync::Iter::new(&client, endpoint);
+
+    for (i, result) in iter.enumerate() {
+        let txn = result?;
+
+        println!("ID:             {}", txn.id());
+        println!("source account: {}", txn.source_account());
+        println!("created at:     {}", txn.created_at());
+        println!("");
+
+        if (i + 1) % 10 == 0 && !utils::next_page() {
+            break;
+        }
+    }
+    Ok(())
+}
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -3,13 +3,33 @@ extern crate stellar_client;
 extern crate stellar_resources;
 
 use clap::{App, AppSettings, Arg, SubCommand};
-use stellar_client::sync::Client;
+use stellar_client::{error::Error, sync::Client};
 
 fn build_app<'a, 'b>() -> App<'a, 'b> {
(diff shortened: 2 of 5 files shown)


cli/src/main.rs
```diff
@@ -39,128 +59,91 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
...
-                ::std::process::exit(1);
-            }
-            err => {
-                eprintln!("An unknown error occurred: {:?}", err);
-                ::std::process::exit(1);
-            }
-        }
-    }
+    let client = if let Some(host) = matches.value_of("host") {
+        Client::new(&host).expect("Failed to initialize client")
```
correlator: 🔥

cli/src/main.rs
```diff
@@ -3,13 +3,33 @@ extern crate stellar_client;
...
         .version("0.1")
         .about("Access the stellar horizon API via the command line.")
         .setting(AppSettings::SubcommandRequired)
+        .arg(
+            Arg::with_name("host")
+                .takes_value(true)
+                .short("h")
+                .long("host")
+                .conflicts_with_all(&["test-net", "pub-net"])
+                .help("The host url to attempt to connect to"),
```
correlator: Is it worth mentioning that the default is test-net here as well.

cli/src/account.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    println!("ID:       {}", account.id());
+    println!("Sequence: {}", account.sequence());
+
+    Ok(())
+}
+
+pub fn transactions<'a>(client: Client, matches: &'a ArgMatches) -> Result<()> {
+    let id = matches.value_of("ID").expect("ID is required");
+    let endpoint = account::Transactions::new(id).order(Order::Desc);
```
correlator: Do we want limit, order, or cursor to be options?

cli/src/account.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    for (i, result) in iter.enumerate() {
+        let txn = result?;
+
+        println!("ID:             {}", txn.id());
+        println!("source account: {}", txn.source_account());
+        println!("created at:     {}", txn.created_at());
+        println!("");
+
+        if (i + 1) % 10 == 0 && !utils::next_page() {
```
correlator: This whole thing is so good.  Especially with the new iterators on Result.

cli/src/main.rs
```diff
@@ -39,128 +59,91 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
...
+    } else if matches.is_present("pub-net") {
+        Client::horizon().unwrap()
+    } else {
+        Client::horizon_test().unwrap()
+    };
 
-    pub fn transactions<'a>(client: Client, matches: &'a ArgMatches, cursor: Option<String>) {
-        let id = matches.value_of("ID").expect("ID is required");
-        let mut endpoint = account::Transactions::new(id).order(Order::Desc);
+    // Master match block. All subcommands need to be captured here.
```
robertDurst: Wow match statements are awesome!

cli/src/account.rs
kbacha: perhaps. Order makes sense, cursor could be a starts-from, and limit is page size (which may not control how many we actually fetch at a time). However, that can be another branch as I think this one contains enough.

PR #110, opened 2018-04-03
Implement effects for accounts: 

david-pm:
https://www.stellar.org/developers/horizon/reference/endpoints/effects-for-account.html

radical

### DIFF
--- a/client/src/endpoint/account.rs
+++ b/client/src/endpoint/account.rs
@@ -1,7 +1,7 @@
 //! Contains endpoints for accessing accounts and related information.
 use error::Result;
 use std::str::FromStr;
-use stellar_resources::{Account, Datum, Transaction};
+use stellar_resources::{Account, Datum, Effect, Transaction};
 use super::{Body, Cursor, IntoRequest, Order, Records};
 use http::{Request, Uri};
 
@@ -94,6 +94,7 @@ mod details_tests {
         assert_eq!(request.uri().path(), "/accounts/abc123/data/key");
     }
 }
+
 /// An endpoint that accesses the transactions for a specific account
 #[derive(Debug, Clone)]
 pub struct Transactions {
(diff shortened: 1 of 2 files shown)


client/src/endpoint/account.rs
```diff
@@ -267,3 +268,111 @@ mod transactions_tests {
...
+    pub fn new(id: &str) -> Self {
+        Self {
+            id: id.to_string(),
+            cursor: None,
+            order: None,
+            limit: None,
+        }
+    }
+
+    /// Starts the page of results at a given cursor
```
kbacha: We've been adding doc-tests to these. Would you mind adding those in?

david-pm: I can pretty much just mirror what we are doing with the transactions one right?

bringking: @david-pm you can see some examples here - https://github.com/kbacha/stellar-sdk/blob/master/client/src/endpoint/transaction.rs#L34

kbacha: Yep, cause you'll need an account id.

kbacha:
@david-pm Don't forget to run `cargo fmt` on your changes 😄

bringking:
@kbacha we should setup a pre-commit hook that does `cargo fmt` for you or something.

kbacha:
@bringking that would either add a bunch of useless commits or it would be local only. If someone wants a pre-commit hook locally, they can do that.

correlator:
Looking pretty sweet brother.  Thanks for joining in!

client/src/endpoint/account.rs
david-pm: no clue what i'm doing haha
i just used the transaction endpoint to get an account_id since i saw us do that earlier but i should be able to get it from the effects endpoint shouldn't i?

bringking: @david-pm yeah I think this is good/right. Basically for these tests, the only way to be sure you get a valid account ID without hardcoding it would be to query for transactions and get the account ID from the transaction.

correlator:
💖

PR #116, opened 2018-04-03
Implements the Payments for Ledger endpoint

### LINKED ISSUE #86: [endpoint] Payments for ledger
This endpoint represents all payment operations that are part of a valid transactions in a given ledger.

```rust
use stellar_client::endpoint::ledger::Payments;
```

https://www.stellar.org/developers/horizon/reference/endpoints/payments-for-ledger.html

kbacha:
Implements the payments route for ledgers. In doing so I implemented
several other changes to facilitate it:

1. Add clone + cursor to the All endpoint for ledgers
2. Remove Deserialize and add Clone to all Operation types
3. Migrate from returning &Sting to returning &str on operation getters
4. Added transaction_hash to operation as a common attribute (not
documented but appears in all the responses).

closes #86

### DIFF
--- a/client/src/endpoint/ledger.rs
+++ b/client/src/endpoint/ledger.rs
@@ -1,8 +1,8 @@
 //! Contains the endpoint for all ledgers.
 use error::Result;
 use std::str::FromStr;
-use stellar_resources::Ledger;
-use super::{Body, IntoRequest, Order, Records};
+use stellar_resources::{Ledger, Operation};
+use super::{Body, Cursor, IntoRequest, Order, Records};
 use http::{Request, Uri};
 
 /// Represents the all ledgers end point for the stellar horizon server. The endpoint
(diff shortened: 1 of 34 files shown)


client/src/endpoint/ledger.rs
```diff
@@ -170,38 +203,185 @@ impl IntoRequest for Details {
...
+
+/// Represents the payments for ledger endpoint on the stellar horizon server.
+/// The endpoint will return all the payment for a single ledger in the chain.
+///
+/// <https://www.stellar.org/developers/horizon/reference/endpoints/payments-for-ledger.html>
+///
+/// ## Example
+/// ```
+/// use stellar_client::sync::Client;
+/// use stellar_client::endpoint::{ledger, payment, transaction};
```
kbacha: Btw, I decided to make this one test much more detailed in lieu of the other examples. If ya'll like this then I think this is a pattern to follow?

image

I think it's nice because it demonstrates the relationship between payments, ledgers, and transactions.

correlator: Wow.  Nice.  It is also the only real way to ensure that a request would yield a payment.  I am 100% in favor of this moving forward.

PR #118, opened 2018-04-04
[#78] Adds payments to the account endpoint

### LINKED ISSUE #78: [endpoint] Payments for account
This endpoint responds with a collection of Payment operations where the given account was either the sender or receiver.

```rust
use stellar_client::endpoint::account::Payments;
```

https://www.stellar.org/developers/horizon/reference/endpoints/payments-for-account.html

alexanderreiff:
This implements the `payments` action on the `account` endpoint, returning a series of `Operation` records: https://www.stellar.org/developers/horizon/reference/endpoints/payments-for-account.html

This was largely copy 🍝 from the `transactions` action, so tell me where I'm off base here.

This closes #78

### DIFF
--- a/client/src/endpoint/account.rs
+++ b/client/src/endpoint/account.rs
@@ -5,15 +5,44 @@ use stellar_resources::{Account, Datum, Effect, Operation, Transaction};
 use super::{Body, Cursor, IntoRequest, Order, Records};
 use http::{Request, Uri};
 
-/// An endpoint that accesses a single accounts details.
+/// Represents the account details on the stellar horizon server.
+/// The endpoint will return information relating to a specific account.
+///
+/// <https://www.stellar.org/developers/horizon/reference/endpoints/accounts-single.html>
+///
+/// ## Example
+/// ```
+/// use stellar_client::sync::Client;
+/// use stellar_client::endpoint::{account, transaction};
+///
+/// let client = Client::horizon_test().unwrap();
+///
+/// // Grab transaction and associated account to ensure an account populated with transactions
+/// let transaction_ep = transaction::All::default().limit(1);
+/// let all_txns       = client.request(transaction_ep).unwrap();
+/// let txn            = &all_txns.records()[0];
+/// let account_id     = txn.source_account();
+///
+/// // Now we issue a request for that account's transactions
+/// let endpoint  = account::Details::new(account_id);
+/// let details   = client.request(endpoint).unwrap();
+///
+/// assert_eq!(details.id(), account_id);
+/// ```
 #[derive(Debug)]
 pub struct Details {
     id: String,
 }
 
 impl Details {
-    /// Returns a new end point for account details. Hand this to the client in order to request
-    /// details about an account.
+    /// Creates a new account::Details endpoint struct. Hand this to the client in order to request
+    /// information relating to a specific account
+    ///
+    /// ```
+    /// use stellar_client::endpoint::account;
+    ///
+    /// let details = account::Details::new("abc123");
+    /// ```
     pub fn new(id: &str) -> Self {
         Self { id: id.to_string() }
     }
@@ -29,9 +58,10 @@ impl IntoRequest for Details {
     }
 }
 
-/// An endpoint that returns a single value for a key/vaule pair associated with an account.
+/// Represents the data for account endpoint on the stellar horizon server.
+/// The endpoint will return a single value for a key/value pair associated with an account.
 ///
-/// https://www.stellar.org/developers/horizon/reference/endpoints/data-for-account.html
+/// <https://www.stellar.org/developers/horizon/reference/endpoints/data-for-account.html>
 ///
 /// ## Example
 /// ```
@@ -42,7 +72,7 @@ impl IntoRequest for Details {
 /// let endpoint    = account::Data::new("GATLAI2D7SSH6PE3HXTDPTRM4RE5VRK6HGA63K5EWP75PSANCZRFDNB5", "Food");
 /// let record      = client.request(endpoint).unwrap();
 /// #
-/// # assert_eq!(record.value(), "Pizza"); //
+/// # assert_eq!(record.value(), "Pizza");
 /// ```
 #[derive(Debug)]
 pub struct Data {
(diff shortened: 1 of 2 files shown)


client/src/endpoint/account.rs
```diff
@@ -452,3 +452,180 @@ mod effects_tests {
...
+    /// # let transaction_ep   = transaction::All::default().limit(1);
+    /// # let txns             = client.request(transaction_ep).unwrap();
+    /// # let txn              = &txns.records()[0];
+    /// # let account_id       = txn.source_account();
+    /// # let endpoint         = account::Transactions::new(account_id);
+    /// # let first_page       = client.request(endpoint).unwrap();
+    /// # // grab first page and extract cursor
+    /// # let cursor           = first_page.next_cursor();
+    /// let endpoint = account::Payments::new(account_id).cursor(cursor);
+    /// let records  = client.request(endpoint).unwrap();
```
correlator: Some of these endpoints require a lot of work to test :)  In this case it looks like the cursor is coming from the transactions endpoint and that we are using it to find our place in the payments endpoint.  I'll admit that I'm not sure if the cursor is a global counter that increments all resources and the docs aren't terribly clear on this either https://www.stellar.org/developers/horizon/reference/resources/page.html

Assuming we did get the cursor from payments, there is still no guarantee that the first account returned from the transaction call has 2 payments, so this test may still blink anyhow which is a bummer.  We have been moving towards testing that the URL is properly constructed and testing the actual cursor functionality on the `All` endpoint.  https://github.com/kbacha/stellar-sdk/blob/master/client/src/endpoint/ledger.rs#L624-L626  I think this should be fine here.

Sorry for the long stream of consciousness response :)

client/src/endpoint/account.rs
```diff
@@ -452,3 +452,180 @@ mod effects_tests {
...
+    /// use stellar_client::sync::Client;
+    /// use stellar_client::endpoint::account;
+    ///
+    /// # use stellar_client::endpoint::transaction;
+    /// let client   = Client::horizon_test().unwrap();
+    /// # let transaction_ep   = transaction::All::default().limit(1);
+    /// # let txns             = client.request(transaction_ep).unwrap();
+    /// # let txn              = &txns.records()[0];
+    /// # let account_id       = txn.source_account();
+    /// let endpoint = account::Payments::new(account_id).limit(1);
```
correlator: Same issue as above with no guarantee that the endpoint could return more than 1 item.

client/src/endpoint/account.rs
```diff
@@ -452,3 +452,180 @@ mod effects_tests {
...
+            limit: None,
+        }
+    }
+
+    /// Starts the page of results at a given cursor
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// use stellar_client::sync::Client;
```
kbacha: Since these tests are a bit involved and require a good deal of set up, let's just do it once in the struct documentation itself and just provide examples of setting the values in the function documentation.

client/src/endpoint/account.rs
kbacha: I believe that @correlator is correct, transactions are not guaranteed to have payments (account creation and offers are in transactions but no payment is made i believe). 

You may try using the `payment::All` endpoint to make sure that you locate an actual payment, that would give you guarantees.

alexanderreiff:
Thanks for the feedback guys! I'll get this fixed up 👍

david-pm:
are we missing in `pub use super::account::Payments as ForAccount` payment.rs ?

client/src/endpoint/account.rs
```diff
@@ -452,3 +471,175 @@ mod effects_tests {
...
+/// ```
+/// use stellar_client::sync::Client;
+/// use stellar_client::endpoint::{account, payment};
+///
+/// let client = Client::horizon_test().unwrap();
+///
+/// // Grab payments and associated account to ensure an account with payments
+/// let all_payments = client.request(payment::All::default().limit(1)).unwrap();
+/// let payment      = &all_payments.records()[0];
+/// let account_id   = payment.from();
```
alexanderreiff: @kbacha @correlator I struggled with the best way to get an account ID from the `payment::All` response. The response struct are generic `Operation` types. The associated account IDs appear contained in the `kind` attribute child struct, but not all Operations have some account identifier and they are not keyed consistently across operation types from what I can grok. 

What would be the Rust Way to get the account ID out of these Operations?

alexanderreiff:
Rewrote the informative tests for the new `account::Payments` action, as well as the existing `account::Transactions` action as requested. Went ahead and made similar doc changes for the rest of the actions in the module for consistency.

I have one question about using the `payment::All` action to pluck out an account ID for use in the `account::Payments` test.

Thanks!

client/src/endpoint/account.rs
```diff
@@ -452,3 +471,175 @@ mod effects_tests {
...
+        let payments = Payments::new("abc123");
+        let req = payments
+            .into_request("https://horizon-testnet.stellar.org")
+            .unwrap();
+        assert_eq!(req.uri().path(), "/accounts/abc123/payments");
+        assert_eq!(req.uri().query(), None);
+    }
+
+    #[test]
+    fn it_can_make_a_payments_uri() {
```
kbacha: Is there any significant difference between this and the above test?

client/src/endpoint/account.rs
kbacha: Something like this may work:

```rust
let account_id = match payment.kind() {
    OperationKind::Payment(payment) => payment.from(),
    _ => panic!("Expected payment type"),
};
```

client/src/endpoint/account.rs
alexanderreiff: Nope. Merged the minor differences between the two.

client/src/endpoint/account.rs
```diff
@@ -617,3 +636,175 @@ mod ledger_operations_tests {
...
+/// use stellar_client::endpoint::{account, payment};
+/// use stellar_resources::operation::OperationKind;
+///
+/// # fn main() {
+/// let client = Client::horizon_test().unwrap();
+///
+/// // Grab payments and associated account to ensure an account with payments
+/// let all_payments = client.request(payment::All::default().limit(1)).unwrap();
+/// let payment      = &all_payments.records()[0];
+/// let account_id   = match payment.kind() {
```
alexanderreiff: Getting this `match` block right turned into a nice Rust learning exercise. 

@kbacha can you give this one more thumbs up for appropriate usage before I merge?

PR #120, opened 2018-04-04
Better cli pagination

### LINKED ISSUE #115: Improved pagination
### Expected Behavior
The user can paginate routes that return multiple entries in several ways. This behavior is configurable the same way across all these apis. Here are the needed cases:

```bash
# Change the page size that we "break" at. This does not need
# to change the actual limit, just where we interrupt the scroll
stellar transactions all --page-by 25

# Walk through and show all results without interruption. This should
# conflict with the page-by
stellar transactions all --all
```

Either way, the methodology for pagination should be done once so that we don't need to duplicate the code across all the commands. The configuration should be the same (maybe a function can return the arg?) but we should paginate using a struct that accepts an Iterator (can paginate anything) and uses a closure to feed results.

```rust
Paginator {
    page_size: PageSize,
}

enum PageSize {
    All,
    Size(usize)
}

impl Paginator {
    fn new(page_size: PageSize) -> Self;

    fn paginate<I>(iter: I) -> Result where I: Iterator;
}
```

### Actual Behavior
Pagination is unconfigurable

kbacha:
To simplify the code, I've combined the arguments, matching, and
pagination into a single module called Pager. This enables reuse of the
same code to be used in every matcher. It can also be theoretically used
for any iterator at all.

fixes #115

### DIFF
--- a/cli/src/pager.rs
+++ b/cli/src/pager.rs
@@ -1,0 +1,135 @@
+use clap::{App, Arg, ArgMatches};
+use std::str::FromStr;
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct Pager {
+    size: PageSize,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum PageSize {
+    All,
+    Size(usize),
+}
+
+impl Pager {
+    pub fn add<'a, 'b>(app: App<'a, 'b>) -> App<'a, 'b> {
+        app.arg(
+            Arg::with_name("all")
+                .long("all")
+                .conflicts_with("page-size")
+                .help("Query for all records"),
+        ).arg(
+            Arg::with_name("page-size")
+                .long("page-by")
+                .conflicts_with("all")
+                .takes_value(true)
+                .validator(Self::validator)
+                .help("Query and display in a specific page size. Defaults to 10"),
+        )
+    }
+
+    fn validator(v: String) -> Result<(), String> {
+        usize::from_str(&v)
+            .map(|_| ())
+            .map_err(|_| String::from("Page size must be a positive integer"))
+    }
+
+    pub fn from_arg<'a>(arg: &'a ArgMatches) -> Pager {
+        if let Some(size) = arg.value_of("page-size") {
+            Pager {
+                size: PageSize::Size(usize::from_str(&size).unwrap_or(10)),
+            }
+        } else if arg.is_present("all") {
+            Pager {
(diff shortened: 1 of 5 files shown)


cli/src/pager.rs
```diff
@@ -0,0 +1,135 @@
...
+    where
+        I: Iterator<Item = T>,
+        F: FnMut(T),
+    {
+        match self.size {
+            PageSize::All => iter.for_each(f),
+            PageSize::Size(page_size) => for (i, item) in iter.enumerate() {
+                f(item);
+
+                if (i + 1) % page_size == 0 && !Self::next_page() {
```
correlator: It appears that when users call `All` they will not have the option to interactively paginate.  Namely `next_page()` is never referenced in the all arm.  Is this intentional or am I missing something?

cli/src/pager.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn validator(v: String) -> Result<(), String> {
+        usize::from_str(&v)
+            .map(|_| ())
+            .map_err(|_| String::from("Page size must be a positive integer"))
+    }
+
+    pub fn from_arg<'a>(arg: &'a ArgMatches) -> Pager {
+        if let Some(size) = arg.value_of("page-size") {
+            Pager {
+                size: PageSize::Size(usize::from_str(&size).unwrap_or(10)),
```
correlator: There is the possibility that people pass large values in.  Is there an advantage to imposing a max on the size of `limit` vs validating that the value is < 200 on initialization?

cli/src/pager.rs
kbacha: That is correct. `all` means `all`... you get the firehose.

cli/src/pager.rs
kbacha: limit is the size of the request to horizon, I'm not worried about people asking for too much.

kbacha: I suppose i could rename the `limit` method to `horizon_page_limit`

cli/src/pager.rs
correlator: well, you get 200 and don't have the option to request more.  Is that correct?

kbacha: No, we request 200 at a time from horizon. You are displayed a fire hose.

correlator: ah right, the iterator on endpoint does this for us.  Could be annoying if I called All on something like effect and just had an endless firehose, but I could also see this as being pretty useful in other contexts.  Cool.  Thanks for going through this with me.

PR #125, opened 2018-04-06
Implement effects for transactions: https://www.stellar.org/developer…

david-pm:
https://www.stellar.org/developers/horizon/reference/endpoints/effects-for-transaction.html

seinfeld

### DIFF
--- a/client/src/endpoint/account.rs
+++ b/client/src/endpoint/account.rs
@@ -410,11 +410,11 @@ impl IntoRequest for Effects {
     }
 }
 
-// impl Cursor<Effect> for Effects {
-//     fn cursor(self, cursor: &str) -> Self {
-//         self.cursor(cursor)
-//     }
-// }
+impl Cursor<Effect> for Effects {
+    fn cursor(self, cursor: &str) -> Self {
+        self.cursor(cursor)
+    }
+}
 
 #[cfg(test)]
 mod effects_tests {
--- a/client/src/endpoint/transaction.rs
+++ b/client/src/endpoint/transaction.rs
@@ -1,7 +1,7 @@
 //! Contains endpoints for transactions and related information.
 use error::Result;
 use std::str::FromStr;
-use stellar_resources::{Operation, Transaction};
+use stellar_resources::{Effect, Operation, Transaction};
 use super::{Body, Cursor, IntoRequest, Order, Records};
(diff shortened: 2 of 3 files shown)


client/src/endpoint/account.rs
```diff
@@ -410,11 +410,11 @@ impl IntoRequest for Effects {
     }
 }
 
-// impl Cursor<Effect> for Effects {
-//     fn cursor(self, cursor: &str) -> Self {
-//         self.cursor(cursor)
-//     }
-// }
```
david-pm: this was commented out copy pasta... do we need it?

kbacha: Yes, the trait implementation is necessary for using the endpoint in an iterator.

client/src/endpoint/transaction.rs
```diff
@@ -162,6 +162,188 @@ mod all_transactions_test {
...
+            limit: None,
+        }
+    }
+
+    /// Starts the page of results at a given cursor
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// use stellar_client::sync::Client;
```
kbacha: We've just started moving the example that makes queries to horizon to just the struct definition itself. This way we provide a large example for clarity and reduce the number of http calls we make.

Here's an example of the new approach:
https://github.com/kbacha/stellar-sdk/blob/master/client/src/endpoint/account.rs#L456-L558

(note: we haven't converted the old examples yet)

Here's what the "new" way looks like in `cargo doc`:

image

client/src/endpoint/transaction.rs
```diff
@@ -162,6 +162,188 @@ mod all_transactions_test {
...
+        let effects = Effects::new("abc123");
+        let req = effects
+            .into_request("https://horizon-testnet.stellar.org")
+            .unwrap();
+        assert_eq!(req.uri().path(), "/transactions/abc123/effects");
+        assert_eq!(req.uri().query(), None);
+    }
+
+    #[test]
+    fn it_can_make_a_actions_uri() {
```
kbacha: How is this different than the previous test?

kbacha:
@david-pm Can you also add a `closes #123` to your description so that when this merges we automatically close the issue?

david-pm:
gonna open another PR

PR #127, opened 2018-04-07
Add all trades endpoint

### LINKED ISSUE #89: [endpoint] All Trades
People on the Stellar network can make offers to buy or sell assets. When an offer is fully or partially fulfilled, a trade happens.

It looks like there are a large number of params but we should be able to accept an `AssetIdentifier` as an argument the base/counter.

```rust
use stellar_client::endpoint::trade::All;
```

https://www.stellar.org/developers/horizon/reference/endpoints/trades.html

robertDurst:
Ok so this is v1... feedback plz. Things to note:

1. For some of the doc tests I need to extern crate. Based on [this gist](https://gist.github.com/killercup/2431b7861f671d20ee00e43d71d8d65c) I wrote a couple doc tests using a main fn.

2. I created a few functions on AssetIdentifier as noted in my discussion with @kbacha in #89. These allow for type safe AssetIdentifier creation whereas in the new function used by asset's deserialize the asset_code takes a string.

3. I lowercased all price order params (it was uppercased before). Whoever wrote this @kbacha or @correlator, if there was a reason for this please let me know. The reason I lowercased the N and D was because that was how it came back from Horizon.

4. Tests will fail because at the moment, specifying the `base_asset` param causes a 500 error to be returned -- I will let the devs know but right now it is 12:24 on a Friday night so I don't think anyone is online right now 😅 

Ref #112 

*Is there a GIF that reflects how this work made you feel?*

dj khaled salsa gif-downsized_large

Closes #89


client/src/endpoint/trades.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// ```
+    /// # extern crate stellar_resources;
+    /// # extern crate stellar_client;
+    /// 
+    /// use stellar_client::sync::Client;
+    /// use stellar_client::endpoint::trades;
+    /// use stellar_resources::AssetIdentifier;
+    /// 
+    /// # fn main() {
+    ///     let base_asset = AssetIdentifier::alphanum4("MOBI", "GA6HCMBLTZS5VYYBCATRBRZ3BZJMAFUDKYYF6AH6MVCMGWMRDNSWJPIH");
```
kbacha: Is this asset guaranteed to exist even if the testnet is refreshed? We've actually started leaving these integration tests out of the function-specific doctests recently actually.

client/src/endpoint/trades.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// ```
+    /// # extern crate stellar_resources;
+    /// # extern crate stellar_client;
+    /// 
+    /// use stellar_client::sync::Client;
+    /// use stellar_client::endpoint::trades;
+    /// use stellar_resources::AssetIdentifier;
+    /// 
+    /// # fn main() {
+    ///     let counter_asset = AssetIdentifier::native();
```
kbacha: You probably should remove the leading spaces so that it doesn't look oddly indented in the docs.

client/src/endpoint/trades.rs
```diff
@@ -0,0 +1,257 @@
...
+        let ep = All::default();
+        let req = ep.into_request("https://www.google.com").unwrap();
+        assert_eq!(req.uri().path(), "/trades");
+        assert_eq!(req.uri().query(), None);
+    }
+
+    #[test]
+    fn it_puts_the_query_params_on_the_uri() {
+        let ep = All::default()
+            .base_asset(AssetIdentifier::native())
```
kbacha: I think that we should have at least two tests here. I'd like to see us test both the native and alpha's for each asset.

resources/fixtures/trade.json
```diff
@@ -24,7 +24,7 @@
   "counter_asset_issuer": "GCKA6K5PCQ6PNF5RQBF7PQDJWRHO6UOGFMRLK3DYHDOI244V47XKQ4GP",
   "base_is_seller": true,
   "price": {
-    "N": 10,
-    "D": 61
+    "n": 10,
```
kbacha: I think we had it upper case because the documentation here:
https://www.stellar.org/developers/horizon/reference/endpoints/trades.html

The example json has uppercases 🤷‍♂️

resources/src/asset.rs
```diff
@@ -140,6 +140,31 @@ impl AssetIdentifier {
             _ => Err("Invalid Asset Type.".to_string()),
         }
     }
+
+    /// A type safe way of creating a native asset AssetIdentifier
+    pub fn native() -> AssetIdentifier {
```
kbacha: We should add some simple tests for these. They are super basic but also nice to have.

resources/fixtures/trade.json
robertDurst: Hmmm that is weird.

See here: https://www.stellar.org/laboratory/#explorer?resource=trades&endpoint=all&values=eyJiYXNlX2Fzc2V0Ijp7InR5cGUiOiIiLCJjb2RlIjoiIiwiaXNzdWVyIjoiIn0sImNvdW50ZXJfYXNzZXQiOnsidHlwZSI6Im5hdGl2ZSIsImNvZGUiOiIiLCJpc3N1ZXIiOiIifX0%3D&network=test

kbacha: Yea, wouldn't be the first time we've seen the response structure be different in the API versus in the documentation.

robertDurst: It'd be awesome if we note these :)

robertDurst:
@kbacha cool cool changes made. If it looks good, lmk so I can rebase.

resources/fixtures/trade.json
kbacha: Absolutely we should note these. Where is the best place to do that? Is the documentation in a public repo?

kbacha:
Generally looks good but it looks like travis is still upset. Also, you should rebase and squash your commits down.

robertDurst:
Cool, will rebase and cleanup. Here is the bug... for some reason specifying the base asset always returns a 500: https://www.stellar.org/laboratory/#explorer?resource=trades&endpoint=all&values=eyJiYXNlX2Fzc2V0Ijp7InR5cGUiOiJuYXRpdmUiLCJjb2RlIjoiIiwiaXNzdWVyIjoiIn19&network=test

resources/fixtures/trade.json
robertDurst: I think it might be, I'll give it a look tomorrow at work

robertDurst:
Wow not sure what I have been doing up to this point, but rebasing with:

```
git fetch upstream
git rebase upstream/master
```

is way easier 😅

robertDurst:
@kbacha only failing test is the 500 horizon error. Otherwise, rebased and other tests pass. Thoughts? 

And I will ask about this tomorrow to see if it is a bug

robertDurst:
Ok, here is the scoop. You have to specify both the counter and the base. 

The issue here, the reason I see 500, is the error message is not hooked up yet.

I will fix this up today.

kbacha:
@robertDurst So, either both are None or both are Some? So a second struct may make sense.

```rust
struct BaseAndCounter {
    base: AssetIdentifier,
    counter: AssetIdentifier,
}

struct All {
    asset: Option<BaseAndCounter>,
    ...
}
```

A tuple might also work but this is a little more descriptive of which is which. As a tuple it would look like:

```rust
struct All {
    base_and_counter: Option<(AssetIdentifier, AssetIdentifier)>,
    ...
}
```

If the field is private it might not even matter. But I think we should capture the requirement in the type itself.

kbacha:
@robertDurst How can I help get this fixed up and merged?

robertDurst:
@kbacha This will be fixed by EOD (or maybe end of night 😄 ) sorry life got crazy

robertDurst:
BOOOM! Checks pass -- let's get @alexanderreiff unblocked

client/src/endpoint/trades.rs
```diff
@@ -0,0 +1,211 @@
...
+    ///
+    /// # fn main() {
+    /// let base_asset = AssetIdentifier::native();
+    /// let counter_asset = AssetIdentifier::native();
+    ///
+    /// let client      = Client::horizon_test().unwrap();
+    /// let endpoint    = trades::All::default().trade_pair(base_asset, counter_asset);
+    /// let records     = client.request(endpoint).unwrap();
+    ///
+    /// # assert_eq!(records.records().len(), 0);
```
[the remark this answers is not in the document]
robertDurst: Here I check for a trade pair that is native, native. While this is ridiculous and obviously returns 0, I am basing this off of @kbacha's recommendation that I move away from tests that may fail in case of test net resets. **Not quite related to what you're re recommending here, just a note.**

robertDurst:
@correlator changes made. 

Last couple things:

What're your thoughts on the docs I have for offer_id and trade_pair. Are these still necessary? I see that we are deviating away from these type of comments.

Also, is this a kosher way to do a private struct?

client/src/endpoint/trades.rs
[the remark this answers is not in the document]
correlator: I thought using XLM was a cool idea that I had not thought of. But given that your tests pass and that you are asserting there is at least one trade with this pairing, it seems like people are trading xlm for xlm 🤷‍♀️ 

If you want to get deep on this you can call 

```
trades::All::default().limit(1)
```
get the asset pair off of that result, then call 
```
trades::All::default().with_trade_pair().limit(1)
``` 
using those assets and assert the returned trade has the correct asset pair.

correlator:
`TradePair` works fine where it is but I'm happy letting Kevin have a look before we merge.  In the meantime, it may be more clear to call it `AssetPair` as it holds a pair of assets rather than a pair of trades.  

Not applicable to this PR but this did get me thinking about how many resources we have that contain a pair of assets (Offers, OrderBooks, etc) and whether it might make sense to introduce a struct like this in the resource crate.

On the topic of not hitting the API with doc-tests that implement filtering, we have been moving in that direction, but we are far from uniform here (there will be a future issue to make the repo doc-tests uniform after we finish the endpoints).  If you would like to delete the tests and avoid all that gymnastics we did above, go for it.  If you find it helpful to write the tests out explicitly here, that's fine with me too, but again happy to wait for other opinions.

PR #129, opened 2018-04-07
Updated the docs a bit

kbacha:
Added some more verbiage to the README and further clarified what kind
of doctests we are looking for. Also adds a few documentation headers
to existing modules to make it a little easier for new developers.

dancing tube man

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,14 +14,20 @@ they can be built together and tested at once.
 ## Resources
 
 The resources crate documents all the various data types in horizon. These resources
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -47,6 +53,9 @@ You can then use it with:
 stellar --help
 ```
 
+The CLI is most useful for easily perusing horizon/stellar data without needing to construct
+curl commands.
```
correlator: Can also be useful as a working example of how to consume the sdk

PR #131, opened 2018-04-07
Add Cursor to the CLI

### LINKED ISSUE #113: Allow pagination to start at a cursor
### Expected Behavior
```
# Starts the set at a specific cursor.
stellar transactions all --cursor ABC123
# or
stellar transactions all --start-at ABC123
```

### Actual Behavior
Cannot specify where to start.

correlator:
Added the cursor argument to the clap app by adding a cursor option
to the listable macro defined in main.  If an argument is supplied
to the cursor it is added to the endpoint before querying the
Horizon API.

Resolves #113 

body positivity gif-downsized

### DIFF
--- a/cli/src/account.rs
+++ b/cli/src/account.rs
@@ -1,6 +1,6 @@
 use clap::ArgMatches;
 use stellar_client::{sync, endpoint::account, error::Result, sync::Client};
-use super::{ordering, pager::Pager};
+use super::{cursor, ordering, pager::Pager};
 
 pub fn details<'a>(client: Client, matches: &'a ArgMatches) -> Result<()> {
     let id = matches.value_of("ID").expect("ID is required");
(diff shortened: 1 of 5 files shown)


cli/src/account.rs
```diff
@@ -17,9 +17,12 @@ pub fn transactions<'a>(client: Client, matches: &'a ArgMatches) -> Result<()> {
     let pager = Pager::from_arg(&matches);
 
     let id = matches.value_of("ID").expect("ID is required");
-    let endpoint = account::Transactions::new(id)
-        .order(ordering::from_arg(matches))
-        .limit(pager.horizon_page_limit() as u32);
+    let endpoint = {
+        let mut endpoint = account::Transactions::new(id)
```
kbacha: Does this actually need to be mutable?

correlator: I actually don't think so.  Same with transactions.  I'll touch this up

PR #134, opened 2018-04-09
Implements a custom derive for cursor

kbacha:
The implementations of cursor (and others) feel a little tedious and
repetitive. They all have the same basic idea and we can simplify their
creation and cut down on extraneous code by implementating a custom
derive. The derivation must live in a separate crate (as I have put it)
and can only expose the proc macro functions. This crate is used to
preprocess and expand the derive tags in the code during the expansion
compilation phase.

We can implement similar traits for Order and Limit. At the same time we
may want to make sure that the trait names are more imperative and
possibly use a better build function name like with_*. These refactors
will come in subsequent changes.


derives/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+
+use proc_macro::TokenStream;
+use syn::DeriveInput;
+
+#[proc_macro_derive(Cursor)]
+pub fn cursor(input: TokenStream) -> TokenStream {
+    // Parse the string representation
+    let ast: DeriveInput = syn::parse(input).unwrap();
+    let name = &ast.ident;
+    let code = quote! {
```
alexanderreiff: This is sweet! :candy:

PR #138, opened 2018-04-10
Add offers to account endpoint

### LINKED ISSUE #60: [endpoint] Offers for Account
People on the Stellar network can make offers to buy or sell assets. This endpoint represents all the offers a particular account makes.

This should go inside a module for accounts.

https://www.stellar.org/developers/horizon/reference/endpoints/offers-for-account.html

alexanderreiff:
As a means to introspect the offers which an account makes, this
implements the account endpoint action for offers.

Fixes #60

Note that the doc-tests here rely on the `trade::All` endpoint being implemented in #127. That will need get merged ahead of this.

### DIFF
--- a/client/src/endpoint/account.rs
+++ b/client/src/endpoint/account.rs
@@ -1,7 +1,7 @@
 //! Contains endpoints for accessing accounts and related information.
 use error::Result;
 use std::str::FromStr;
-use stellar_resources::{Account, Datum, Effect, Operation, Transaction};
+use stellar_resources::{Account, Datum, Effect, Offer, Operation, Transaction};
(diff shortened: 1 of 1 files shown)


kbacha:
FYI, at this point, there will soon be a Limit trait and Order trait that can be derived. Should simplify even more 👍 

It also appears there's a compile err?
https://travis-ci.org/kbacha/stellar-sdk/builds/365400419?utm_source=github_status&utm_medium=notification

alexanderreiff:
Updated to use `Limit` and `Order` traits and addressed the silly compile error (made a tiny name change one place but not the order 🤦‍♂️ ).

This still has one failing doc test until #127 merges.

PR #140, opened 2018-04-10
Resolves all 130 clippy issues on resources

kbacha:
Mostly these were issues around simplied struct creation. I did make one
major change to the way thresholds are passed into set set_options
operation.

mr-clean


correlator:
what level of clippy did you use?  clippy, clippy_style, or clippy_pedantic?

kbacha:
@correlator I used the default... I didn't specify anything.

PR #143, opened 2018-04-11
Add an order trait and derive it

### LINKED ISSUE #135: Procedural macro for order
Any endpoint that returns a collection of records takes a parameter to specify the order in which records should be returned.  We currently implement this manually on every such endpoint.  The end state of this issue requires 3 tasks to be completed:

- Procedural macro to define a custom derive for Order
- Replace the duplicated code in index endpoints with derive Order
- Add a test to ensure that order is still being set on the endpoint.

kbacha:
To simplify the creation of endpoints and maintenance, this creates an
`Order` trait and renames the direction of the order to `Direction`.
This allows us to then use a procedural macro to derive the trait. Also,
it enables some additional simplifications in the cli in the future.

Also in this change is modifying asset_code and asset_issuer setters to
use the `with_` prefix.

closes #135

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,8 @@ nursery: https://github.com/rust-lang-nursery/rustfmt
 Naming and conventions to improve the public interface should be followed and considered. We want
 sdk to be as easy to use as possible.
 
+### Traits
+
 When developing a trait please consider that traits are more like verbs than nouns. As such the
 convention has been to utilize words like `Clone` instead of `Cloneable` and `Write` instead of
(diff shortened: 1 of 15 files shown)


CONTRIBUTING.md
```diff
@@ -107,6 +109,22 @@ what it does rather than what it is.
...
+```rust
+
+// BAD
+fn set_value(mut self, val: Value) -> Self;
+
+// BAD
+fn value(mut self, val: Value) -> Self;
+
+// GOOD
+fn with_value(mut self, val: Value) -> Self;
```
bringking: Nice addition

PR #144, opened 2018-04-12
Changelog

### LINKED ISSUE #132: Add a CHANGELOG.md
We should keep a changelog for when we actually release. This should keep track of all changes that a developer using our SDK would want to take note of. I like the style of this one:

https://keepachangelog.com/en/1.0.0/

We should probably add it for each crate that we intend to release. Along with the changelog files themselves, we should also note in the contributing file how to update the changelog. PRs that update functionality, or fix things should make note of it in the changelog along with their PR. Once we release, all the changes that are part of the release will be segmented off into the new version.

For the initial release this is not as necessary as we are going from 0 to 1. But this is a requirement for milestone 0.2.0.

kbacha:
Adds the change log with a few of the additions. Every crate should
possibly have it's own so I've put a changelog in each crate's
directory. I've also added a mention in contributing to update the
change log if needed with each PR.

fixes #132 

why-not-both

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,6 +138,12 @@ describe the code change but provide clarity around why it was changed. It's ok 
 locally but when submitting your branch for review, please squash all commits into a logical series
 of changes.
 
+#### Changelogs
+
+We keep a change log of additions and fixes. Please review how changelogs are kept at the [keepachangelog](https://keepachangelog.com/en/1.0.0/)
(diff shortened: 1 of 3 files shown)


correlator:
What are your thoughts on updating this on every dot release vs at the time of changing a feature?

kbacha:
@correlator I think that at least trying to keep it updated for each change will save us time for a release. Not everything will go into it (internal refactors won't matter, for instance) but if we can keep up with it I think our lives will be easier.

PR #151, opened 2018-04-13
DRY code using limit and order traits

kbacha:
Now that there are traits for limit and order, we can reduce the amount
of code needed by using generics to assign them.

### DIFF
--- a/cli/src/ordering.rs
+++ b/cli/src/ordering.rs
@@ -1,5 +1,5 @@
 use clap::{App, Arg, ArgMatches};
-use stellar_client::endpoint::{Direction, Direction::Asc, Direction::Desc};
+use stellar_client::endpoint::{Direction::{Asc, Desc}, Order};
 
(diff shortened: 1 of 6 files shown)


cli/src/ordering.rs
```diff
@@ -19,17 +19,36 @@ pub fn add<'a, 'b>(app: App<'a, 'b>) -> App<'a, 'b> {
 }
 
 /// Parses the argument matches and returns the order to use.
-pub fn from_arg(arg: &ArgMatches) -> Direction {
-    match arg.value_of(ARG_NAME) {
+pub fn assign_from_arg<T>(arg: &ArgMatches, order: T) -> T
+where
+    T: Order,
+{
```
correlator: stronger signature. Nice

kbacha:
It's more "DRY" because it results in fewer method calls. Before I had to get the value from arg and then assign. Now it just assigns. Number of lines may be the same.

PR #174, opened 2018-04-15
Add payments for transaction to CLI

### LINKED ISSUE #149: Add Transaction Payments to the CLI
Query horizon for the payments that resulted from a transaction.  (Payments are operations, so go figure)

### Expected Behavior
```
stellar transaction payments <ID>
```
This operation should also take cursor, order and page-limit as arguments.

### Payment details to display
```
id
to
from
asset
amount
```

You can use the account::transactions cli operation as a template.

alexanderreiff:
Adds a command to request payment operations made as part of a given
transaction to the CLI application.

Fixes #149

### DIFF
--- a/cli/src/transactions.rs
+++ b/cli/src/transactions.rs
@@ -1,4 +1,5 @@
 use stellar_client::{endpoint::{transaction, Limit, Order}, error::Result, sync::{self, Client}};
+use stellar_resources::OperationKind;
 use clap::ArgMatches;
 use super::{cursor, ordering, pager::Pager};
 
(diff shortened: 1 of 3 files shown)


cli/src/transactions.rs
```diff
@@ -24,3 +25,37 @@ pub fn all(client: &Client, matches: &ArgMatches) -> Result<()> {
...
+        cursor::assign_from_arg(matches, endpoint)
+    };
+    let iter = sync::Iter::new(&client, endpoint);
+
+    let mut res = Ok(());
+    pager.paginate(iter, |result| match result {
+        Ok(op) => {
+            println!("ID:       {}", op.id());
+            match op.kind() {
+                &OperationKind::Payment(ref pymt) => {
```
kbacha: Sometimes when only a single match, it's nice to use `if let`

```rust
if let &OperationKind(ref pymt) = op.kind() {
    // stuff
} else {
    panic!
}
```

PR #180, opened 2018-04-18
Add effects for transaction to CLI

### LINKED ISSUE #158: Add Transaction Effects endpoint to CLI
### Expected Behavior
```
stellar transaction effects <ID>
```
The ID is the Transaction's ID

### Effect details to display
```
id
kind
```
https://www.stellar.org/developers/horizon/reference/endpoints/effects-for-transaction.html

alexanderreiff:
Adds a command to request all effects resulting from a given transaction
to the CLI application.

All Effects resources are modified to be Clone types.

This also fixes a copy-paste error in 68265ca

Fixes #158

### DIFF
--- a/cli/CHANGELOG.md
+++ b/cli/CHANGELOG.md
@@ -12,3 +12,5 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Transaction payments can be seen
 - Transaction details can be seen
 - All ledgers can be seen
+- Transaction operations can be seen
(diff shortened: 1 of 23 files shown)


robertDurst:
@alexanderreiff 👍 

@correlator I noticed I had to add the Clone trait as well. I am familiar with clone -- it creates a new instance so that if you have:
```
let a = val;
let b = a.clone()
```

then ```a``` isn't moved. Is this the basic idea here as well?

PR #183, opened 2018-04-19
Add trades endpoint to CLI

### LINKED ISSUE #173: Add Trades Endpoint to CLI
### Expected Behavior
```
stellar trades
```
This should also take optional arguments:
```
- base_asset_type
- base_asset_code
- base_asset_issuer
- counter_asset_type
- counter_asset_code
- counter_asset_issuer
- offer_id 
- cursor
- order
- page-limit
```
### Trade details to display
Each returned trade should show the following details
```
id
offer_id
paging_token
ledger_close_time
base_account
base_amount
base_asset
counter_amount
counter_account
counter_asset
price
seller
```
https://www.stellar.org/developers/horizon/reference/endpoints/trades.html

correlator:
Resolves #173 

In the process also updated the name of the trade endpoint initializer
from with_trade_pair to with_asset_pair as that more accurately describes
what the method does.

https://media.giphy.com/media/Oivne3ir8TzcA/200w_d.gif

### DIFF
--- a/cli/src/trades.rs
+++ b/cli/src/trades.rs
@@ -1,0 +1,68 @@
+use stellar_client::{endpoint::trades, sync::{self, Client}};
+use stellar_resources::AssetIdentifier;
+use clap::ArgMatches;
+use super::{cursor, ordering, pager::Pager};
+use error::Result;
+
+pub fn all(client: &Client, matches: &ArgMatches) -> Result<()> {
+    let pager = Pager::from_arg(&matches);
+
+    let endpoint = {
+        let mut endpoint = trades::All::default();
+
+        if let Some(offer_id) = matches.value_of("offer_id") {
+            let offer_id = offer_id.parse::<u32>().unwrap();
+            endpoint = endpoint.with_offer_id(offer_id);
+        };
+        if let (Some(base_asset_type), Some(counter_asset_type)) = (
+            matches.value_of("base_asset_type"),
+            matches.value_of("counter_asset_type"),
+        ) {
+            let base_asset_code = matches
+                .value_of("base_asset_code")
+                .map(|code| code.to_string());
+            let base_asset_issuer = matches
+                .value_of("base_asset_issuer")
+                .map(|issuer| issuer.to_string());
+            let counter_asset_code = matches
+                .value_of("counter_asset_code")
+                .map(|code| code.to_string());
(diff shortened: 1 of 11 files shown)


cli/src/trades.rs
```diff
@@ -0,0 +1,76 @@
...
+            let counter_asset_code = matches
+                .value_of("counter_asset_code")
+                .map(|code| code.to_string());
+            let counter_asset_issuer = matches
+                .value_of("counter_asset_issuer")
+                .map(|issuer| issuer.to_string());
+            let base_asset =
+                match AssetIdentifier::new(base_asset_type, base_asset_code, base_asset_issuer) {
+                    Ok(asset) => asset,
+                    Err(error) => panic!("There was a with the supplied base_asset: {:?}", error),
```
kbacha: There's no reason to panic here. You should find a way to map this error to the returned Result. Right now we are piggy backing on the errors from the client library. Perhaps we can find a way to wrap that into another error type and allow the errors to flow back so that the program exits in a controlled manner.

robertDurst: @correlator Pardon my interruption into this convo/pr, but I just came across ?'s (the question mark) in a random tangent the other day. Would using a ? here allow the errors to flow back up if this function returned a Result type? Or am I way off talking about something unrelated (very likely)?

correlator: You are absolutely right.  A `?` would allow us to short circuit and return the error and the final implementation of this will likely utilize that operation.  Right now I'm trying to decide which error struct is most appropriate to return.  The CLI currently utilizes errors from the `stellar_client` crate this error doesn't feel like it belongs in `stellar_client`.   I was admittedly cutting corners to panic rather than creating the appropriate error to return.  Going to be digging into this tonight.

robertDurst: Thanks for the response. I'll admit, I copy-pasta a bit much, usually just as a starting point and I noticed similar code with a panic that I started to use -- after testing it a bit, the panic felt like an abrupt end. From my understanding, the only reason ever to panic is to deal with non-code errors, i.e. unpredictable runtime events. Coming from Javascript and 7 days worth of GO so panic and real error handing in general is pretty new.

kbacha: @robertDurst @correlator Yup, it will work very well if the type implements the `From` trait.

PR #191, opened 2018-04-21
Add payments for ledgers to the CLI and edit a few things in the docs

### LINKED ISSUE #170: Add Payments for Ledger to CLI
Query horizon for the payments that resulted from a transaction.  (Payments are operations, so go figure)

### Expected Behavior
```
stellar ledger payments <ID>
```
This operation should also take cursor, order and page-limit as arguments.

### Payment details to display
```
id
to
from
asset
amount
```

https://www.stellar.org/developers/horizon/reference/endpoints/payments-for-ledger.html

correlator:
Resolves #170 

Adds payments for a ledger to the CLI and cleaned up a few things in the documentation.

screen shot 2018-04-21 at 2 12 15 pm

### DIFF
--- a/cli/src/ledgers.rs
+++ b/cli/src/ledgers.rs
@@ -1,4 +1,4 @@
-use stellar_client::{endpoint::ledger, sync::{self, Client}};
+use stellar_client::{endpoint::ledger, resources::OperationKind, sync::{self, Client}};
 use clap::ArgMatches;
 use error::Result;
 use super::{cursor, ordering, pager::Pager};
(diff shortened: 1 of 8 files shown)


cli/src/ledgers.rs
```diff
@@ -27,3 +27,38 @@ pub fn all(client: &Client, matches: &ArgMatches) -> Result<()> {
...
+            println!("ID:           {}", op.id());
+            if let &OperationKind::Payment(ref pymt) = op.kind() {
+                println!("To account:   {}", pymt.to());
+                println!("From account: {}", pymt.from());
+                println!("Asset Type:   {}", pymt.asset().asset_type());
+                println!("Asset Code:   {}", pymt.asset().code());
+                println!("Asset Issuer: {}", pymt.asset().issuer());
+                println!("Amount:       {}", pymt.amount());
+            } else {
+                panic!("Did not receive Payment operation");
```
kbacha: If this ever happens I'll 😱 

Would it matter if we just left this `else` arm off?

cli/src/ledgers.rs
```diff
@@ -27,3 +27,38 @@ pub fn all(client: &Client, matches: &ArgMatches) -> Result<()> {
...
     res
 }
+
+pub fn payments(client: &Client, matches: &ArgMatches) -> Result<()> {
+    let pager = Pager::from_arg(&matches);
+    let sequence = matches
+        .value_of("sequence")
+        .expect("Ledger sequence is required");
+
+    let sequence = sequence.parse::<u32>().unwrap();
```
kbacha: Feels like we can give better feedback to the user than unwrapping this parse.

cli/src/ledgers.rs
correlator: @kbacha Yes we can get rid of the else arm. @robertDurst very interesting.  I can't seem to get a path_payment to be returned from the api, but i definitely believe that it could be.  Will match on the result to do the printing accordingly.  Thanks for flagging this 👍

correlator:
@robertDurst @kbacha have give it another look at your convenience.

PR #198, opened 2018-04-29
Convert ledgers to formatter/render

kbacha:
weekend-warror


robertDurst:
Looks great!

PR #196, opened 2018-04-28
Add all payments to CLI

robertDurst:
So I'll admit there is some nice copy-pasta here -- this is very similar to the ledgers payments cli.

However, I did find an error in the current CLI payments (found in both the ledgers and transactions). If you scrolled through until you found a path_payment, you would see that you get the error:

> Missing fields for path_payment operation.

The reason for this is that `source_amount` does not exist for path payments. See here:

https://www.stellar.org/laboratory/#explorer?resource=transactions&endpoint=single&values=eyJ0cmFuc2FjdGlvbiI6IjAwNzk0ZGUyYmQxYzEyMTliMzUyNjU3YmNlMWU0ZDhjZTJmNjA3MzdiYzU1OGZmM2MwNDZiMTRkZTRhYWExMGQifQ%3D%3D&network=public

This field only exists on payment paths, an object describing an intermediate step on a path_payment.

### Is there a GIF that reflects how this work made you feel?

Closes #167

PR #200, opened 2018-04-29
Scott/all payments

### LINKED ISSUE #172: Add Transaction for Ledger Endpoint to CLI
### Expected Behavior
```
stellar ledger transactions <ID>
```
The ID is the Ledger's ID

### Transaction details to display
```
id
source_account
created_at
```
https://www.stellar.org/developers/horizon/reference/endpoints/transactions-for-ledger.html

correlator:
Added all payments endpoint to the CLI.  As @robertDurst pointed out, source amount is not actually returned as a part of the path_payment operation from horizon so I updated that resource and corresponding models.

Also had some time on a plane so added transactions for ledger to the CLI as well.

Resolves #196  Resolves #172 

Alt Text

### DIFF
--- a/cli/src/fmt/simple/payment.rs
+++ b/cli/src/fmt/simple/payment.rs
@@ -1,0 +1,78 @@
+use fmt::Render;
+use super::Simple;
+use stellar_client::resources::{Operation, OperationKind};
+
+impl Render<Operation> for Simple {
+    fn render(&self, operation: &Operation) -> Option<String> {
+        let mut buf = String::new();
+        append_to_buffer!(buf, "ID:             {}", operation.id());
+        match operation.kind() {
+            &OperationKind::CreateAccount(ref create_account) => {
+                append_to_buffer!(buf, "Operation Kind:   Create Account");
+                append_to_buffer!(buf, "Account:          {}", create_account.account());
+                append_to_buffer!(buf, "Funder:           {}", create_account.funder());
+                append_to_buffer!(
+                    buf,
+                    "Starting Balance: {}",
+                    create_account.starting_balance()
+                );
+            }
+            &OperationKind::Payment(ref payment) => {
+                append_to_buffer!(buf, "Operation Kind: Payment");
+                append_to_buffer!(buf, "To account:     {}", payment.to());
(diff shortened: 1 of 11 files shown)


robertDurst:
This is great -- get rid of the redundancies!

I was curious, is there any way to test this? Ie is there a way to catch the error I found without actually manually running the CLI?

correlator:
Well, there are 2 places where we test this.  The first is in the fixture file that contains our best understanding of what we expect horizon to return https://github.com/kbacha/stellar-sdk/blob/master/client/fixtures/operations/path_payment.json . You can see that is now updated in this PR.  The other place is in the endpoints testing suite.  Problem is I'm not sure how to specifically request a path_payment from the horizon api.

robertDurst:
@correlator that is valid. I believe you'd have to find one in the ledger and then request it specifically -- thus to guarantee all tests are consistent in the future, we'd want to use the public network.

cli/src/fmt/simple/payment.rs
```diff
@@ -0,0 +1,73 @@
...
+                append_to_buffer!(buf, "Account:          {}", create_account.account());
+                append_to_buffer!(buf, "Funder:           {}", create_account.funder());
+                append_to_buffer!(
+                    buf,
+                    "Starting Balance: {}",
+                    create_account.starting_balance()
+                );
+            }
+            &OperationKind::Payment(ref payment) => {
+                append_to_buffer!(buf, "Operation Kind: Payment");
```
kbacha: Feels like we could have the "kind" put first

cli/src/fmt/simple/payment.rs
```diff
@@ -0,0 +1,73 @@
+use fmt::Render;
+use super::Simple;
+use stellar_client::resources::{Operation, OperationKind};
+
+impl Render<Operation> for Simple {
+    fn render(&self, operation: &Operation) -> Option<String> {
+        let mut buf = String::new();
+        append_to_buffer!(buf, "ID:             {}", operation.id());
+        match operation.kind() {
```
kbacha: Clippy would say to match by the deref.

PR #201, opened 2018-04-29
Add operations for account to CLI

### LINKED ISSUE #163: Add Operations for Account Endpoint to CLI
### Expected Behavior
```
stellar account operations <ID>
```
The ID is the Account's ID

### Operation details to display
```
id
kind
```
https://www.stellar.org/developers/horizon/reference/endpoints/operations-for-account.html

### LINKED ISSUE #203: Build out simple formatter for operations in CLI
This is going to be a giant match that prints out all the relevant fields for each operation type.  

https://www.stellar.org/developers/horizon/reference/resources/operation.html

Currently, we only display the id and kind, which is enough to call the operation details url, but showing the fields directly will be a nice UX improvement.

alexanderreiff:
Add operations for account to CLI

Adds a command to request all operations relating to a given account
to the CLI application.

This also adds a formatter for each kind of operation, so the details
of each operation get included with the output.

Fixes #163, fixes #203


cli/src/account.rs
```diff
@@ -34,3 +34,23 @@ pub fn transactions(client: &Client, matches: &ArgMatches) -> Result<()> {
...
+    let endpoint = pager.assign(endpoint);
+    let endpoint = cursor::assign_from_arg(matches, endpoint);
+    let endpoint = ordering::assign_from_arg(matches, endpoint);
+
+    let iter = sync::Iter::new(&client, endpoint);
+
+    let mut res = Ok(());
+    let mut fmt = Formatter::start_stdout(Simple);
+    pager.paginate(iter, |result| match result {
+        Ok(op) => fmt.render(&op),
```
correlator: 😎

cli/src/fmt/simple/operation/mod.rs
```diff
@@ -0,0 +1,12 @@
+use fmt::Render;
```
kbacha: Why put this in a `mod.rs` file versus just an `operation.rs` file?

cli/src/fmt/simple/operation/mod.rs
```diff
@@ -0,0 +1,12 @@
...
+use super::Simple;
+use stellar_client::resources::Operation;
+
+impl Render<Operation> for Simple {
+    fn render(&self, op: &Operation) -> Option<String> {
+        let mut buf = String::new();
+        append_to_buffer!(buf, "ID:   {}", op.id());
+        append_to_buffer!(buf, "Kind: {}", op.kind_name());
```
kbacha: Perhaps we can expand on this with the various details of the operations?

correlator: I added a separate issue for this as it is going to be a big and obnoxious lift.  Feel free to tackle in this PR or we can come back to it later. https://github.com/kbacha/stellar-sdk/issues/203

alexanderreiff: I'm happy to tackle it. Three of them will be taken care of here: https://github.com/kbacha/stellar-sdk/pull/200

cli/src/fmt/simple/operation/mod.rs
alexanderreiff: I envisioned in the future we'd have renders in here for each of the `OperationKind`s

kbacha: I don't think it'll be as long as you think. The effects had a similar size number of variations.

alexanderreiff:
@correlator @kbacha can you guys take another look at this? Thanks!

cli/Cargo.toml
```diff
@@ -11,3 +11,4 @@ path = "src/main.rs"
 chrono = { version = "0.4" }
 clap = "2.31"
 stellar-client = { path = "../client" }
+textwrap = "0.10"
```
correlator: It seems like we are pulling this in to use this method https://github.com/mgeisler/textwrap/blob/master/src/lib.rs#L870-L880 . 

Kevin and I were just talking about indentation so it put a big smile on my face when I saw this, but maybe we can just pull in this one method.  Are there other things in the library we also may want to use?

cli/src/fmt/mod.rs
```diff
@@ -17,6 +22,14 @@ pub trait Render<T> {
         None
     }
 
+    fn render_nested(&self, item: &T) -> Option<String> {
```
correlator: This is so effing awesome!

cli/src/fmt/simple/asset.rs
```diff
@@ -1,21 +1,38 @@
...
+impl Render<Flags> for Simple {
+    fn render(&self, flags: &Flags) -> Option<String> {
+        let mut buf = String::new();
+        if flags.is_auth_required() {
+            append_to_buffer!(buf, "{}", indent("auth is required".to_string()));
         }
-        if asset.is_auth_revocable() {
-            append_to_buffer!(buf, "  auth is revocable");
+        if flags.is_auth_revocable() {
+            append_to_buffer!(buf, "{}", indent("auth is revocable".to_string()));
```
correlator: I <3 the indent so much

cli/src/fmt/simple/mod.rs
```diff
@@ -6,13 +6,21 @@ macro_rules! append_to_buffer {
     }
 }
 
+macro_rules! nest_in_buffer {
```
correlator: 👍

cli/src/fmt/mod.rs
kbacha: I don't think this should be part of the trait. This feels very particular to the simple format and should probably be implemented on the simple type itself.

If it's implemented in `Simple` then we could even store the ident level in it. That would mean that we can increment when we create the inner renderer. Then nesting is aware of the depth.

client/src/resources/operation/allow_trust.rs
```diff
@@ -44,8 +44,17 @@ impl AllowTrust {
...
     }
 
-    /// The limit for the asset.
+    /// `true` when trust allowed; `false` when revoked
     pub fn authorize(&self) -> bool {
         self.authorize
     }
+
+    /// Whether trust has been allowed or revoked
+    pub fn status(&self) -> &str {
```
kbacha: I'm not sure I like this if it's just for display.

cli/src/fmt/simple/asset.rs
```diff
@@ -1,21 +1,38 @@
...
-            append_to_buffer!(buf, "  auth is required");
+        nest_in_buffer!(self, buf, &asset.flags());
+        Some(buf)
+    }
+}
+
+impl Render<AssetIdentifier> for Simple {
+    fn render(&self, asset_id: &AssetIdentifier) -> Option<String> {
+        let mut buf = String::new();
+        append_to_buffer!(buf, "Type:   {}", asset_id.asset_type());
```
kbacha: In another place I just made this `<code>-<issuer` unless native, and then it's just `XLM`. Can we do that here as well?

alexanderreiff: If we want just a one-liner output, should this implement `fmt::Display` rather than `Render`?

alexanderreiff:
@kbacha can you take a look at this refactor? Took me a bit of Rust-y try and error to figure out the indent level–aware nesting, but in the end, the solution I got was pretty simple.

cli/src/fmt/simple/mod.rs
```diff
@@ -6,16 +6,73 @@ macro_rules! append_to_buffer {
...
+const INDENT_SPACES: usize = 4;
+
+pub struct Simple {
+    nest_level: usize,
+}
+
+impl Simple {
+    pub fn new() -> Self {
+        Self { nest_level: 0 }
+    }
```
correlator: 👍

cli/src/fmt/simple/mod.rs
```diff
@@ -6,16 +6,73 @@ macro_rules! append_to_buffer {
...
+        }
+    };
+}
+
+macro_rules! indent_in_buffer {
+    ($buffer:expr, $fmt:ident, $($args:tt)*) => {
+        let nested = $fmt.nest().align_str(&format!($($args)*));
+        $buffer.push_str(&nested);
+    }
+}
```
correlator: This distinction is aweome

cli/src/account.rs
```diff
@@ -20,7 +20,7 @@ pub fn details(client: &Client, matches: &ArgMatches) -> Result<()> {
     let endpoint = account::Details::new(id);
     let account = client.request(endpoint)?;
 
-    Formatter::start_stdout(Simple).render(&account);
+    Formatter::start_stdout(Simple::new()).render(&account);
```
kbacha: I'm kicking myself for not creating a `new` function from the get-go.

cli/src/fmt/simple/operation.rs
```diff
@@ -0,0 +1,173 @@
+use super::Simple;
+use fmt::Render;
+use stellar_client::resources::{OperationKind as Kind, operation::*};
+
+impl Render<Operation> for Simple {
+    fn render(&self, op: &Operation) -> Option<String> {
+        let mut buf = String::new();
+        append_to_buffer!(buf, "ID:   {}", op.id());
+        append_to_buffer!(buf, "Kind: {}", op.kind_name());
+        let kind_details = match *op.kind() {
```
kbacha: with rust 1.26 you don't need to deref or add `ref` to the kinds.

cli/src/fmt/simple/mod.rs
```diff
@@ -6,16 +6,73 @@ macro_rules! append_to_buffer {
     }
 }
 
-pub struct Simple;
+macro_rules! nest_in_buffer {
```
kbacha: I wonder if we can remove the `_in_buffer` and `_to_buffer` so that it's more like a dsl and just has 3 verbs: `append!`, `indent!`, and `nest!`.

I believe it's properly scoped to only the simple module so it shouldn't harm any other modules.

alexanderreiff: I like this a lot better! I was finding the `_in_buffer` naming awkward.

PR #207, opened 2018-05-02
Add orderbook details to CLI and make a killer formatter

### LINKED ISSUE #166: Add Orderbook Details Endpoint to CLI
### Expected Behavior
```
stellar orderbook details <ID>
```

### Orderbook details to display
```
bids
asks
base
counter
```
https://www.stellar.org/developers/horizon/reference/endpoints/orderbook-details.html

correlator:
Resolves #166 

Alt Text

### DIFF
--- a/cli/src/asset_identifier.rs
+++ b/cli/src/asset_identifier.rs
@@ -1,0 +1,50 @@
+use error::InvalidInputError;
+use stellar_client::resources::AssetIdentifier;
+
+pub fn from_str(s: &str) -> Result<AssetIdentifier, InvalidInputError> {
+    let tokens: Vec<&str> = s.split('-').collect();
+    if tokens.len() > 2 {
+        return Err(InvalidInputError::from_str(
+            "Asset identifier not of the form <asset_code>-<asset_issuer>",
+        ));
+    }
+
+    match (tokens.get(0), tokens.get(1)) {
+        (Some(&"XLM"), None) => return Ok(AssetIdentifier::Native),
+        (Some(code), Some(issuer)) if code.len() <= 4 => {
+            return Ok(AssetIdentifier::alphanum4(code, issuer))
+        }
+        (Some(code), Some(issuer)) if code.len() <= 12 => {
+            return Ok(AssetIdentifier::alphanum12(code, issuer))
+        }
+        _ => {
+            return Err(InvalidInputError::from_str(
+                "Asset identifier not of the form <asset_code>-<asset_issuer>",
+            ))
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn it_can_parse_asset_identifiers() {
+        assert_eq!(from_str("XLM").unwrap(), AssetIdentifier::Native);
+        assert_eq!(
+            from_str("fox-123ABC").unwrap(),
+            AssetIdentifier::alphanum4("fox", "123ABC")
+        );
+        assert_eq!(
+            from_str("starfox-123ABC").unwrap(),
+            AssetIdentifier::alphanum12("starfox", "123ABC")
+        );
+    }
+
+    #[test]
+    fn it_returns_appropriate_errors() {
+        assert!(from_str("fox-123-abs").is_err());
+        assert!(from_str("foxisareallycoolanimal-123").is_err());
+    }
+}
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -21,6 +21,14 @@ pub struct InvalidInputError {
     details: String,
 }
 
+impl InvalidInputError {
+    pub fn from_str(details: &str) -> InvalidInputError {
+        InvalidInputError {
+            details: details.to_string(),
+        }
+    }
+}
+
 impl fmt::Display for InvalidInputError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{}", self.details)
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -11,6 +11,7 @@ use error::CliError;
 
 mod account;
 mod assets;
+mod asset_identifier;
 mod cursor;
 mod effects;
 mod error;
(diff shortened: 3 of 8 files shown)


cli/src/orderbook.rs
```diff
@@ -0,0 +1,42 @@
+use clap::ArgMatches;
+use error::Result;
+use fmt::{Formatter, Simple};
+use stellar_client::{endpoint::{orderbook, Limit}, resources::AssetIdentifier, sync::Client};
+
+pub fn details(client: &Client, matches: &ArgMatches) -> Result<()> {
+    let endpoint = {
+        let base_asset_type = matches
```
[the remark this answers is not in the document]
correlator: a little `AssetIdentifier::from_matches` love sounds like a good time

cli/src/asset_identifier.rs
```diff
@@ -0,0 +1,66 @@
+use error::InvalidInputError;
+use stellar_client::resources::AssetIdentifier;
+
+pub fn parse_identifier_from_str(s: &str) -> Result<AssetIdentifier, InvalidInputError> {
+    let asset_parts: Vec<&str> = s.split("-").collect();
+    if asset_parts[0] == "xlm" {
```
kbacha: Are identifiers case insensitive? Cause I think that lumens are "XLM".

cli/src/asset_identifier.rs
```diff
@@ -0,0 +1,66 @@
...
+            "credit_alphanum4",
+            Some(asset_parts[0].to_string()),
+            Some(asset_parts[1].to_string()),
+        ).map_err(|e| InvalidInputError { details: e });
+    } else if asset_parts[0].len() < 13 {
+        return AssetIdentifier::new(
+            "credit_alphanum12",
+            Some(asset_parts[0].to_string()),
+            Some(asset_parts[1].to_string()),
+        ).map_err(|e| InvalidInputError { details: e });
```
kbacha: have you tried `.into()`??

cli/src/error.rs
```diff
@@ -18,7 +18,7 @@ pub enum CliError {
 /// Errors resulting from incomplete or invalid user input
 #[derive(Debug)]
 pub struct InvalidInputError {
-    details: String,
+    pub details: String,
```
kbacha: According to the `From` implementations I don't think you should have to expose this.

cli/src/asset_identifier.rs
```diff
@@ -0,0 +1,66 @@
+use error::InvalidInputError;
+use stellar_client::resources::AssetIdentifier;
+
+pub fn parse_identifier_from_str(s: &str) -> Result<AssetIdentifier, InvalidInputError> {
+    let asset_parts: Vec<&str> = s.split("-").collect();
```
kbacha: use `splitn` so that we don't split the issuer.

cli/src/asset_identifier.rs
```diff
@@ -0,0 +1,66 @@
...
+    if asset_parts[0] == "xlm" {
+        return Ok(AssetIdentifier::Native);
+    }
+    if asset_parts.len() != 2 {
+        return Err(InvalidInputError {
+            details: "Asset identifier not of the form <asset_code>-<asset_issuer>".to_string(),
+        });
+    }
+    if asset_parts[0].len() < 5 {
+        return AssetIdentifier::new(
```
kbacha: I think there are easier to use constructors than this now:

https://github.com/kbacha/stellar-sdk/blob/master/client/src/resources/asset.rs#L145-L163

cli/src/asset_identifier.rs
```diff
@@ -0,0 +1,66 @@
+use error::InvalidInputError;
+use stellar_client::resources::AssetIdentifier;
+
+pub fn parse_identifier_from_str(s: &str) -> Result<AssetIdentifier, InvalidInputError> {
+    let asset_parts: Vec<&str> = s.split("-").collect();
+    if asset_parts[0] == "xlm" {
+        return Ok(AssetIdentifier::Native);
+    }
+    if asset_parts.len() != 2 {
```
kbacha: You may try using pattern matching for this https://play.rust-lang.org/?gist=85d7266724a99db847d6741730ddebe6&version=stable&mode=debug

correlator: *reacts with 🔥 emoji*

cli/src/asset_identifier.rs
correlator: Won't work if they put too many `-`'s in as splitn returns the rest of the string once it hit's the nth occurrence.  We could still check the last string to see if there is an extra `-` but no gain over current implementation and I like this one.  But thanks for pointing out splitn.  For sure good to know about.

cli/src/error.rs
correlator: It keeps saying it's private.  Is there something I'm doing wrong or missing here?  If not, I'll write a from_str() method.

cli/src/main.rs
```diff
@@ -283,40 +314,18 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
...
-                                    .takes_value(true)
-                                    .help("Filters trades with a counter_asset_code.  Not required for XLM"),
+                                    .required(true)
+                                    .help("Filters trades with a given base_asset. format:  <asset_code>-<asset_issuer>, or xlm if lumens"),
                             )
                             .arg(
-                                Arg::with_name("counter_asset_issuer")
-                                    .long("counter_asset_issuer")
+                                Arg::with_name("counter_asset")
+                                    .long("counter_asset")
```
kbacha: Do we really need to type `_asset` everytime we use this? I feel like just `base` and `counter` might be slightly easier to use on the command line.

cli/src/orderbook.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub fn details(client: &Client, matches: &ArgMatches) -> Result<()> {
+    let endpoint = {
+        let base_asset_str = matches
+            .value_of("base_asset")
+            .expect("Base asset is a required field");
+        let counter_asset_str = matches
+            .value_of("counter_asset")
+            .expect("Counter asset is a required field");
+        let base_asset = asset_identifier::parse_identifier_from_str(base_asset_str)?;
```
kbacha: this is really long: `asset_identifier::parse_identifier_from_str`

I think we should consider that if we are using this from the module, the name of the method can probably just be `from_str`:

```rust
let base = asset_identifier::from_str(base_str)?;
```

cli/src/orderbook.rs
```diff
@@ -0,0 +1,30 @@
...
+        let counter_asset = asset_identifier::parse_identifier_from_str(counter_asset_str)?;
+        let mut endpoint = orderbook::Details::for_asset_pair(base_asset, counter_asset);
+        if let Some(limit) = matches.value_of("limit") {
+            let limit = limit
+                .parse::<u32>()
+                .map_err(|_| String::from("Limitshould be a valid u32 integer"))?;
```
kbacha: y'know, now that you have a constructor for the Invalid error type `from_str` you can probably use it for the return values.

PR #208, opened 2018-05-03
Add Memo field to Transaction Resource

robertDurst:
### What's Going on here?
I realized we were missing the memo field. The obvious beginning was to use an intermediate transaction struct as @correlator elegantly recommended, in order to:

> slurp everything up in the intermediate step and the derive[deserialize] will automagically turn json into the intermediate representation

This went alright until I ran into the issue where the memo content could be:
* Some(String)
* Some(i64)
* None()

For this I experimented quite a bit and eventually settled with a combination of the enum strategy [outlined here](https://users.rust-lang.org/t/deserializing-enum-json-field-with-serde-with-multiple-types/5424) and the [serde default value for a field](https://serde.rs/attr-default.html) for a custom deserialization. WOW crazy stuff.

### What to criteque:
* My comments may be weak or lacking-- please let me know, I wrote these at 1:30 AM.
* Is this the best testing strategy? I create a new fixture for each memo type, copying and pasting the rest of the json structure along with it.
* My types get really messy with references to enums inside Options. `Looking for recommendations` on ways to refactor.
* Still working on good rust style. 

Also, rust fmt isn't currently working, I will get that figured out again before the final PR.


client/src/resources/transaction.rs
```diff
@@ -1,12 +1,117 @@
...
+    Id(MemoValue),
+    Hash(MemoValue),
+    Return(MemoValue),
+    None(),
+}
+
+impl Memo {
+    /// The type of the memo for this transaction
+    pub fn memo_type(&self) -> &str {
+        match self {
```
kbacha: favor `match *self` so you don't need all the references.

client/src/resources/transaction.rs
```diff
@@ -1,12 +1,117 @@
...
+            &Memo::Text(ref memo) => Some(memo),
+            &Memo::Id(ref memo) => Some(memo),
+            &Memo::Hash(ref memo) => Some(memo),
+            &Memo::Return(ref memo) => Some(memo),
+            &Memo::None() => None,
+        }
+    }
+
+    /// Creates and returns a new memo from the given type and MemoValue
+    pub fn new(memo_type: &str, memo: MemoValue) -> Result<Memo, String> {
```
kbacha: I don't believe this needs to be public.

client/src/resources/transaction.rs
```diff
@@ -99,6 +236,16 @@ impl Transaction {
     pub fn fee_meta_xdr(&self) -> &String {
         &self.fee_meta_xdr
     }
+
+    /// The value/contents of the memo for this transaction
+    pub fn memo(&self) -> Option<&MemoValue>{
```
kbacha: I'd rather this just return `Option<&Memo>` as an enum type with the values in the middle.

client/src/resources/transaction.rs
```diff
@@ -1,12 +1,117 @@
...
+    }
+
+    /// The value/contents of the memo for this transaction
+    pub fn memo(&self) -> Option<&MemoValue> {
+        match self {
+            &Memo::Text(ref memo) => Some(memo),
+            &Memo::Id(ref memo) => Some(memo),
+            &Memo::Hash(ref memo) => Some(memo),
+            &Memo::Return(ref memo) => Some(memo),
+            &Memo::None() => None,
```
kbacha: If there's gonna be `Option<Memo>` then we shouldn't need `None` on this type.

client/src/resources/transaction.rs
```diff
@@ -1,12 +1,117 @@
...
+/// Memo values represent the three types of data
+/// that may make up the contents of a memo:
+/// a. Some(String)
+/// b. Some(i64)
+/// c. None()
+#[derive(Deserialize, Debug, Clone, PartialEq, Eq,)]
+pub enum MemoValue {
+    SomeString(String),
+    SomeI64(i64),
+    None()
```
kbacha: Like the memo itself, if the value is none, should it even have a memo at all?

kbacha:
Great work, this clearly has a lot of thought put into it.

kbacha:
Which type of memo needed to be one or the other?

robertDurst:
The memoID is type i64 but not my memo().memo() it returns string

dtolnay:
Based on the test data is only `Memo::Id` allowed to hold a number? That would make things easier.

```rust
#[derive(Deserialize, Debug)]
struct Transaction {
    id: String,

    /* other fields */

    #[serde(flatten)]
    memo: Memo,
}

#[derive(Deserialize, Debug)]
#[serde(rename_all = "lowercase", tag = "memo_type", content = "memo")]
enum Memo {
    Text(String),
    Id(i64),
    Hash(String),
    Return(String),
    None,
}
```

robertDurst:
@dtolnay thanks a lot for the insight! I have incorporated the above into my latest PR. Will need to now go back and see if ```#[serde(flatten)]``` can be used anywhere else 😄  

@kbacha still working the types so this is dev friendly. What is your opinion of:

```transaction.memo().get_type() // &str```

and 

```transaction.memo().get_value() // Option<String>```.

I think these are nice helper functions... or should we leave the matching and unpacking to the user?

**Comment by dtolnay** [comment]
Can you write out an example of code that would want to call `get_type` and get a &str?

**Comment by robertDurst** [comment]
I would probably call the transaction payment endpoint seen [here](https://docs.rs/stellar-client/0.1.1/stellar_client/endpoint/transaction/struct.Payments.html). Note: I have not really used the `map` and `filter` methods (except for the similar methods in JS), so while I assume this works, this is more of a conceptual example.
```rust
// Receive the transaction from the api call
let transaction = payments.records();

/// check the type of the memo
payments.records()
    .map(|p| p.transaction())
    .filter(|t| t.memo().memo_type() == "text")

/// Then I would check for certain memo values of this certain memo_type
/// Ex: I am waiting for my aunt to send me a payment with the memo robs_money
```

robertDurst:
As per my discussion with @kbacha this has been nicely simplified. I removed the methods on the memo enum. Thus, this is how you would access the values of the Memo:
```rust
if let Memo::Text(value) = txn.memo() {

}
```
and of course you can use the enum variants as types.

client/src/resources/transaction.rs
```diff
@@ -1,11 +1,26 @@
 use chrono::prelude::*;
 use super::{deserialize, amount::Amount};
 
+/// Memos are a useful source for adding meta data to a transaction.
+/// A consists of a type and content (unless memo type is none).
+///
+/// To learn more about the concept of memos in the Stellar network, take a look at the Stellar memo description here:
+/// https://www.stellar.org/developers/guides/concepts/transactions.html#memo
```
correlator: I think if you wrap the url in <> it will auto-link in the docs.

robertDurst: fixed!

robertDurst:
@kbacha @correlator rebased to update after last PR merge. Since you both approved this, is this good to go -- can it be merged?

client/src/resources/transaction.rs
```diff
@@ -162,6 +204,55 @@ mod transaction_tests {
...
+            assert_eq!(value, 19292920);
+        }
+    }
+
+    #[test]
+    fn it_parses_memo_return() {
+        let transaction: Transaction =
+            serde_json::from_str(&transaction_json_memo_return()).unwrap();
+
+        if let &Memo::Return(ref value) = transaction.memo() {
```
kbacha: testing tip: When you use a conditional that should be true, you might consider putting a panic in the `else` branch so that if a different type is returned it fails.

robertDurst: Awesome yes, if the if-let case fails the test will still pass. Code updated!

PR #212, opened 2018-05-10
Run cargo fmt for v1.26

kbacha:
Now that there's a new version, there's a small change to the formatter.
This updates the code base for that.

One thing it appears to do differently is that the order of `use` is changes to put `super` and `self` first.

fresh-and-clean-2


cli/src/main.rs
```diff
@@ -29,12 +29,8 @@ mod transactions;
 fn build_app<'a, 'b>() -> App<'a, 'b> {
     macro_rules! listable {
         ($e:expr) => {
-            Pager::add(
-                ordering::add(
-                    cursor::add($e)
-                )
-            )
-        }
+            Pager::add(ordering::add(cursor::add($e)))
```
correlator: hmmm...I like the old formatting better...Oh well.

client/src/endpoint/asset.rs
```diff
@@ -1,9 +1,9 @@
 //! Contains endpoints for assets and related information to specific assets.
-use error::Result;
-use std::str::FromStr;
-use resources::Asset;
 use super::{Body, Cursor, Direction, IntoRequest, Limit, Order, Records};
+use error::Result;
 use http::{Request, Uri};
+use resources::Asset;
+use std::str::FromStr;
```
correlator: ordering makes me so happy

client/src/resources/operation/test.rs
```diff
@@ -14,12 +14,12 @@ mod errors_on_missing_fields_for_type {
...
-                        "type_i":{},
-                        "type":"{}",
-                        "transaction_hash":"123"
-                    }}"#,
+                                    "id":"1",
+                                    "paging_token":"7",
+                                    "type_i":{},
+                                    "type":"{}",
+                                    "transaction_hash":"123"
+                                }}"#,
```
correlator: This section seems to be tripping travis up

PR #215, opened 2018-05-13
Add path payment (find path) endpoint to cli

### LINKED ISSUE #168: Add Find Path Endpoint to CLI
### Expected Behavior
```
stellar find_path <SOURCE_ID> <DESTINATION_ID> <DESTINATION_ASSET_TYPE>  <DESTINATION_AMOUNT>
```
This should optionally take destination_asset_code and destination_asset_issuer if the asset is not native (XLM).

### Operation details to display
If not paths are found:
```
No Paths Found
```
if 1 or more paths are found
```
path0[]
path1[]
path2[]
...
```
https://www.stellar.org/developers/horizon/reference/endpoints/path-finding.html

robertDurst:

It's a maze path

A few things to especially note, ponder, and critique: 
* [Here](https://github.com/kbacha/stellar-sdk/compare/master...robertDurst:master#diff-89a40a34ad4635149f3d3bf5e5bb0f49R41) I have to translate the user input into the 10 ^ 7 we use for the Amount type. Sort of weird IMO.
* I do this [iterator peekable thing here](https://github.com/kbacha/stellar-sdk/compare/master...robertDurst:master#diff-89a40a34ad4635149f3d3bf5e5bb0f49R47) and then check to see if the [iterator is empty here](https://github.com/kbacha/stellar-sdk/compare/master...robertDurst:master#diff-89a40a34ad4635149f3d3bf5e5bb0f49R51) since it is possible for the user to receive 0 records in return. If this is the case I print: `No results found. Please try again.`
* I call this cli subcommand `find-path`. Should this be an underscore or is a dash fine (I based this off of pub_net vs pub-net -- we use the dash for pub-net)?
* Again, wrote these comments when I was pretty tired so if the comments are weak, please let me know 😅 

Closes #168


cli/src/find_path.rs
```diff
@@ -0,0 +1,62 @@
...
+            destination_asset_type,
+            destination_asset_code,
+            destination_asset_issuer,
+        )?;
+
+        let endpoint = payment::FindPath::new(
+            source_account,
+            destination_account,
+            destination_asset,
+            Amount::new(destination_amount.parse::<i64>().unwrap() * 10000000),
```
correlator: I can understand why this doesn't make you feel good.  We ultimately decided that because stellar works with stroops and displays units scaled down by 10^7 to its users, we would follow the same convention. https://www.stellar.org/developers/guides/concepts/assets.html#amount-precision-and-representation . 

I would recommend adding underscores to that big number where a comma would typically go i.e. `10_000_000`

cli/src/main.rs
```diff
@@ -143,6 +144,46 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
...
+                                    .takes_value(true)
+                                    .help("The asset issuer account id for the destination asset"),
+                            )
+                            .arg(
+                                Arg::with_name("destination_asset_code")
+                                    .long("destination_asset_code")
+                                    .takes_value(true)
+                                    .help("The code of the destination asset"),
+                            )
+                            .arg(
```
correlator: We've added a new asset_identifier struct to the CLI https://github.com/kbacha/stellar-sdk/blob/master/cli/src/asset_identifier.rs and we can now shorten the asset inputs like so https://github.com/kbacha/stellar-sdk/blob/master/cli/src/main.rs#L175-L180

cli/src/fmt/simple/payment_path.rs
```diff
@@ -0,0 +1,65 @@
...
+            payment_path.source_asset().asset_type()
+        );
+        append_to_buffer!(
+            buf,
+            "Source asset issuer:              {}",
+            payment_path.source_asset().issuer()
+        );
+        append_to_buffer!(buf, "PATH:");
+        for asset in payment_path.path().iter() {
+            append_to_buffer!(buf, "   Path asset code:                  {}", asset.code());
```
correlator: Looks like you have some leading spaces here to indent the path attributes.  I can't wait to have @alexanderreiff 's pr merged so you can utilize all the awesome tooling he's putting out around this.

cli/src/find_path.rs
```diff
@@ -0,0 +1,62 @@
...
+        endpoint
+    };
+
+    let mut iter = sync::Iter::new(&client, endpoint).peekable();
+
+    let mut res = Ok(());
+
+    if !&iter.peek().is_some() {
+        println!("No results found. Please try again.")
+    }
```
correlator: I think that printing nothing on a get that has no results is fine, but happy to defer to kevin here.

cli/src/fmt/simple/payment_path.rs
robertDurst: Dude pumped here too. I'll leave this here for now.

cli/src/find_path.rs
robertDurst: Ok, Ill go with nothing -- will allow me to get rid of this weird peekable stuff.

cli/src/main.rs
robertDurst: Ok sweet, noted. Got so wrapped up in work and that memo PR from last week I missed some of these :fire: changes.

robertDurst:
Thanks for the review Scott. I say we put this on hold until after @alexanderreiff 's PR is merged so that I can reformat/adjust my indentation for the payment paths.

@correlator one thing I noticed, and this is from @kbacha so maybe he has thoughts on it, the assetidentifier CLI is 🔥 , however it requires you to do "XLM" -- it is case sensitive. Would y'all be open to matching any casing aka "xlm" or "XLM"?

I am sure the answer is probably -- yeah sure go fix it 😄 but just wanted to hear your thoughts in case you think differently.

alexanderreiff:
@robertDurst the indent/nest stuff is finally merged! The macro names are now simplified slightly at @kbacha's suggestion: https://github.com/kbacha/stellar-sdk/blob/37764c033a75cb1bb1054e52f26e90daa5045d7d/cli/src/fmt/simple/mod.rs#L2-L23

robertDurst:
Awesome! Updated for @alexanderreiff 's sweet formatting improvement. 

screen shot 2018-05-14 at 8 28 22 pm

cli/src/fmt/simple/payment_path.rs
```diff
@@ -0,0 +1,76 @@
...
+        let mut buf = String::new();
+        append!(
+            buf,
+            "Destination amount:               {}",
+            payment_path.destination_amount()
+        );
+        append!(
+            buf,
+            "Destination asset code:           {}",
+            payment_path.destination_asset().code()
```
alexanderreiff: We have a `Render` implementation for `AssetIdentifier` now that gives the "standard" 1 liner: `XLM` or `<code>-<issuer>`. You could call `self.render(payment_path.destination_asset()).unwrap()` and not have to append each part individually.

robertDurst: Nice catch! Updated

cli/src/fmt/simple/payment_path.rs
```diff
@@ -0,0 +1,38 @@
...
+            self.render(payment_path.destination_asset()).unwrap()
+        );
+        append!(
+            buf,
+            "Source amount:                    {}",
+            payment_path.source_amount()
+        );
+        append!(
+            buf,
+            "Source asset:                     {}",
```
kbacha: I'm not sure that we need so much space between label and value here.

cli/src/main.rs
```diff
@@ -155,6 +156,38 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
...
+                                Arg::with_name("source_account")
+                                    .long("source_account")
+                                    .takes_value(true)
+                                    .required(true)
+                                    .help("The source account id of the payment path to query"),
+                            )
+                            .arg(
+                                Arg::with_name("destination_asset")
```
kbacha: There's only one asset that needs to be specified. It might make sense to label the action `asset` instead of making it longer.

In fact, you may think about how we can make this easy to use from the command line. Long argument names are a pain to type out and so it might make sense to make them shorter (the name of the arg can stay the same):

```
stellar find-path --from=<source account id> --to=<destination account id> --asset=XLM --amount=1.123456
```

cli/src/find_path.rs
```diff
@@ -0,0 +1,46 @@
...
+            .expect("Destination amount is a required field");
+        let destination_asset_str = matches
+            .value_of("destination_asset")
+            .expect("Destination asset is a required field");
+        let destination_asset = asset_identifier::from_str(destination_asset_str)?;
+        let endpoint = payment::FindPath::new(
+            source_account,
+            destination_account,
+            destination_asset,
+            Amount::new(destination_amount.parse::<i64>().unwrap() * 10_000_000),
```
kbacha: Parsing to an `i64` will lose the decimals: `1.123` will be `1`, etc. We should try to capture what they mean exactly, similar to how we do it in the client library. In fact, it might make sense to even implement a `FromStr` trait on the amount type so that we can handle the parsing cleanly (and without fidelity loss).

robertDurst: Ok, so this was fun, but a little wonky -- I noted in the code w/ comments where the wonkiness comes into play.

robertDurst:
@kbacha thanks for the feedback. Made the appropriate changes. Let me know if there is anything else, or if my `FromStr` or `From` trait implementations (is that the correct terminology?) can be done better.

Also note, a trade_aggregations test started to fail so I just extended the start_time parameter to encompass a trade.

cli/src/find_path.rs
```diff
@@ -0,0 +1,47 @@
...
+            .expect("Destination amount is a required field");
+        let destination_asset_str = matches
+            .value_of("asset")
+            .expect("Destination asset is a required field");
+        let destination_asset = asset_identifier::from_str(destination_asset_str)?;
+        let endpoint = payment::FindPath::new(
+            source_account,
+            destination_account,
+            destination_asset,
+            Amount::from_str(destination_amount).unwrap(),
```
kbacha: I generally like to use `expect` here instead so that you can provide a message. Of course, we should really be returning an Result from this path and allowing it to be handled by the outer logic.

client/src/endpoint/payment.rs
```diff
@@ -235,6 +235,24 @@ impl IntoRequest for FindPath {
     }
 }
 
+impl TryFromUri for FindPath {
+    fn try_from_wrap(wrap: &UriWrap) -> ::std::result::Result<FindPath, uri::Error> {
+        let params = wrap.params();
+        let destination_asset = AssetIdentifier::new(
+            params.get_ok("destination_asset_type")?,
+            params.get_parse("destination_asset_code").ok(),
```
kbacha: These are just strings right? Do you need to get parse them?

client/src/endpoint/payment.rs
```diff
@@ -235,6 +235,24 @@ impl IntoRequest for FindPath {
...
+            params.get_parse("destination_asset_issuer").ok(),
+        )?;
+
+        Ok(FindPath {
+            source_account: params.get_parse("source_account")?,
+            destination_account: params.get_parse("destination_account")?,
+            destination_asset: destination_asset,
+            destination_amount: Amount::new(params.get_parse("destination_account")?),
```
kbacha: Technically if `Amount` implements the `FromStr` trait, you should be able to parse to it.

client/src/resources/amount.rs
```diff
@@ -197,3 +198,86 @@ mod deserialize_amount_tests {
         assert!(amount.is_err());
     }
 }
+
+/// Recreating ParseIntError since its interior is private
+/// https://doc.rust-lang.org/src/core/num/mod.rs.html#3949
+#[derive(Debug)]
+pub enum ParseIntError {
```
kbacha: Better would be to use `ParseAmountError` since that's what we are parsing.

client/src/resources/amount.rs
```diff
@@ -197,3 +198,86 @@ mod deserialize_amount_tests {
...
+
+// There is a built in __description() method that gives
+// the error as a string, however that is only availible on
+// nightly...
+// So I went with @amosbatto's hack here:
+// https://github.com/rust-lang/rust/issues/22639
+impl From<num::ParseIntError> for ParseIntError {
+    fn from(error: num::ParseIntError) -> Self {
+        match error.to_string().as_ref() {
+            "cannot parse integer from empty string" => ParseIntError::Empty,
```
kbacha: These may change with different versions of rust so I wouldn't rely on them for matching if possible. If you want to convert I would actually just wrap the ParseIntError:

```rust
pub enum ParseAmountError {
    IntError(ParseIntError),
}
```

Then the from implementation is easier:

```rust
impl From<num::ParseIntError> for ParseAmountError {
    fn from(error: num::ParseIntError) -> ParseAmountError {
        ParseAmountError::IntError(error)
    }
}
```

If desired, you can implement `Error` trait on this if you want to output the inner description. But It should capture the data you need.

client/src/resources/amount.rs
```diff
@@ -197,3 +198,86 @@ mod deserialize_amount_tests {
...
+            _ => panic!("Unexpected error variant!"),
+        }
+    }
+}
+
+impl FromStr for Amount {
+    type Err = ParseIntError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.rfind('.') {
```
kbacha: This is copy-pasta from the deserialize impl, I think. It would be better to switch to using this implementation to implement the Deserialize function and map to the correct errors that it expects there.

client/src/resources/amount.rs
```diff
@@ -197,3 +198,86 @@ mod deserialize_amount_tests {
...
+    }
+}
+
+#[cfg(test)]
+mod from_str_amount_tests {
+    use super::*;
+
+    #[test]
+    fn it_raises_amount_by_ten_million() {
+        let amount: Amount = Amount::from_str("2.12").unwrap();
```
kbacha: another way:
```rust
let amount: Amount = "2.12".parse().unwrap();
```

cli/src/find_path.rs
robertDurst: Would this be more appropriate -- to parse and throw an error as part of the argument matching process? Although I admit the double expect does not seem right.

```rust
let destination_amount: Amount = matches
            .value_of("amount")
            .expect("Destination amount is a required field")
            .parse()
            .expect("Amount must be properly formatted");
```

This would then allow:
```rust
let endpoint = payment::FindPath::new(
            source_account,
            destination_account,
            destination_asset,
            destination_amount,
        );
```

client/src/resources/amount.rs
robertDurst: Yes, I admit I copy-pasta this. To clarify, are you saying I should keep FromStr and then utilize the FromStr implementation as part of deserialize, removing the code duplication, then mapping the errors of FromStr to appropriately match the Deserialize error type?

client/src/resources/amount.rs
robertDurst: Oh yes, thanks good point. I guess the reason I created the entire matching was because I was unable to figure out how to properly throw an error here:

```rust
if number_decimals > 7 {
                    Err(ParseIntError::Underflow)
                }
```

The reason is that the ParseIntError's value is private so I couldn't instantiate/create a new ParseIntError.

client/src/endpoint/payment.rs
robertDurst: Yes, true. Maybe I am missing something, but when I do:

```rust
params.get("destination_asset_code")
```

I get the following error:
```
note: expected type `std::option::Option<std::string::String>`
               found type `std::option::Option<&str>`
```

client/src/endpoint/payment.rs
robertDurst: Cool, makes sense. However (noob question alert 😄) I get this error:

the trait `std::convert::From<resources::amount::ParseIntError>` is not implemented for `uri::Error`

meaning the two error types are different. Here and in general, how do I deal with situations where a functions expected error type and an expressions error type differ?

robertDurst:
@kbacha thanks for the review and thanks for your patience! I am struggling more so with this PR than I thought I would -- although I am having a pretty good time working with traits and what not.

cli/src/find_path.rs
kbacha: Yup, that would be better I think. I've been recommending just returning an `Error` so that exit is centralized, but it's a CLI so it's not the end of the world.

Eventually, centralized error reporting, would provide a consistently formatted error text though.

client/src/endpoint/payment.rs
kbacha: Ah, it wants it to be owned, so you are actually parsing it. Carry on, nothing to see here.

client/src/endpoint/payment.rs
kbacha: So, that would mean that we are missing a conversion for our error type for amount into the result error type.

Also, I just noticed that you are parsing the account into an amount. That can't be right.

robertDurst: 🤔 that is a valid point

robertDurst:
@kbacha so I made all the above edits except I still don't use the `from_str` trait for:

``` destination_amount: Amount::new(params.get_parse("destination_account")?)```

Unless I am mistaken, I could use `params.get` which would return a `&str`. However `from_str` wants a `String`, so it may be cleaner to just use `get_parse`?

Must say the deserialize looks pretty awesome:
```rust
impl<'de> Deserialize<'de> for Amount {
    fn deserialize<D>(d: D) -> Result<Amount, D::Error>
    where
        D: Deserializer<'de>,
    {
        let s = String::deserialize(d)?;
        let amount: Amount = Amount::from_str(&s).map_err(|e| match e {
            ParseAmountError::ExceedsDecimalLength => {
                de::Error::custom("Amount has too many digits of precision.")
            }
            ParseAmountError::IntError(err) => de::Error::custom(err.to_string()),
        })?;
        Ok(amount)
    }
}
```

Map error with a match is pretty nifty -- assuming this is correct use of map_err + match 😄

robertDurst:
This latest commit may have gone a bit deep down the rabbit hole 😅 

Essentially, I needed to implement `TryFromUri` for the find path endpoint. Here is what I got to work, which is pretty nice:
```rust
impl TryFromUri for FindPath {
    fn try_from_wrap(wrap: &UriWrap) -> ::std::result::Result<FindPath, uri::Error> {
        let params = wrap.params();
        Ok(FindPath {
            source_account: params.get_parse("from")?,
            destination_account: params.get_parse("to")?,
            destination_asset: params.get_parse("asset")?,
            destination_amount: params.get_parse("amount")?,
        })
    }
}
```

To do this however, I had to do a few things:
* add `ParseAmountError` and `ParseAssetError` as enums on the `Uri::Error`
* ported the asset_identifier from_str to Asset Resource

I think this all logically is sound, however I am not 100% sure this is the best approach.

client/src/resources/amount.rs
```diff
@@ -141,23 +142,82 @@ impl<'de> Deserialize<'de> for Amount {
...
+    }
+}
+
+/// Recreating ParseIntError since its interior is private
+/// https://doc.rust-lang.org/src/core/num/mod.rs.html#3949
+#[derive(Debug)]
+pub enum ParseAmountError {
+    /// An error describing the case where the amount contains
+    /// too many digits of precision to parse correctly
+    ExceedsDecimalLength(String),
```
kbacha: I don't think we need a string inside this type. I think the type explains what's wrong well enough.

client/src/resources/amount.rs
```diff
@@ -141,23 +142,82 @@ impl<'de> Deserialize<'de> for Amount {
         D: Deserializer<'de>,
     {
         let s = String::deserialize(d)?;
+        let amount: Amount = Amount::from_str(&s).map_err(|e| match e {
+            ParseAmountError::ExceedsDecimalLength(_val) => {
```
kbacha: Hopefully we can remove the string inner value since it's ignored anyhow.

client/src/resources/asset.rs
```diff
@@ -253,6 +255,97 @@ mod asset_identifier_tests {
     }
 }
 
+/// When a bad token or string is provided to parsing into an asset
+/// you get an error.
+#[derive(Debug)]
+pub struct ParseAssetError {
+    details: String,
```
kbacha: Strings make for bad error typing. You should use an Enum with the possible errors coded into it. Especially as there is only one error possible.

client/src/resources/asset.rs
```diff
@@ -253,6 +255,97 @@ mod asset_identifier_tests {
...
+    }
+}
+
+impl fmt::Display for ParseAssetError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.details)
+    }
+}
+
+impl FromStr for AssetIdentifier {
```
kbacha: I see we are adding this trait impl here but I don't see it removed from the CLI, is there a reason for that?

client/src/resources/asset.rs
```diff
@@ -253,6 +255,97 @@ mod asset_identifier_tests {
     }
 }
 
+/// When a bad token or string is provided to parsing into an asset
+/// you get an error.
+#[derive(Debug)]
+pub struct ParseAssetError {
```
kbacha: This error type is confusing. We are parsing the asset identifier, not an asset itself, right? If so, this is named incorrectly.

client/src/endpoint/trade.rs
```diff
@@ -359,7 +359,7 @@ mod all_trades_tests {
 ///
 /// // Place the start time in the past so we capture it.
 /// let agg = trade::Aggregations::new(base, counter)
-///     .with_start_time(now - 3_000_000_000)
+///     .with_start_time(now - 50_000_000_000)
```
correlator: This is always going to be falling out of style over time.  Should we just start at 0?

client/src/resources/asset.rs
robertDurst: Will hold off for a future PR -- I should probably create a new issue

client/src/endpoint/trade.rs
robertDurst: 🤔

client/src/resources/operation/test.rs
```diff
@@ -14,12 +14,12 @@ mod errors_on_missing_fields_for_type {
             fn $type_to_check() {
                 let json = format!(
                     r#"{{
-                        "id":"1",
-                        "paging_token":"7",
-                        "type_i":{},
-                        "type":"{}",
-                        "transaction_hash":"123"
-                    }}"#,
+                                    "id":"1",
```
kbacha: Looks like you are on an old version of rust now, try updating so this doesn't change.

robertDurst: oh man... let me check.

robertDurst: @kbacha just ran cargo fmt --all w/ 1.26.2 installed and nothing changed. However, moved it back and ran fmt and still nothing changed 🤔 . Anyway just reformatted and pushed.

robertDurst:
@kbacha thoughts?

client/src/resources/amount.rs
```diff
@@ -141,23 +142,80 @@ impl<'de> Deserialize<'de> for Amount {
...
+    }
+
+    #[test]
+    fn it_errors_floats_with_more_than_7_decimals() {
+        let amount = serde_json::from_str::<Amount>("\"0.212847948\"");
+        assert!(amount.is_err());
+    }
+}
+
+/// Recreating ParseIntError since its interior is private
```
kbacha: comment should be changed to reflect changes.

cli/src/main.rs
```diff
@@ -178,6 +179,38 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
                     )
                 ),
         )
+        .subcommand(
+            SubCommand::with_name("find-path")
+                 .about("Fetch possible payment paths")
+                            .arg(
```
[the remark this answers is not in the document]
robertDurst: lol again, not caught (and at the same time not fixed) by cargo fmt

robertDurst:
@kbacha changes made!

robertDurst:
Once this is good to go, will rebase + squash

robertDurst:
Rebased and Squashed.

Need a new: 

found

PR #216, opened 2018-05-14
Set some aliases for lumens

kbacha:
box in box


robertDurst:
Awesome, thanks!

PR #224, opened 2018-06-17
Remove FromStr for AssetIdentifier from CLI

robertDurst:
Pikachu dance

Removed the duplicate FromStr from `cli` since it is now in `client` as of #215.

In doing this, I used the `.parse()`, `.map_err()` strategy which we use in many other places throughout `cli`. For example:
```rust
if let Some(offer_id) = matches.value_of("offer_id") {
            let offer_id = offer_id
                .parse::<u32>()
                .map_err(|_| String::from("Offer Id should be a valid u32 integer"))?;
            endpoint = endpoint.with_offer_id(offer_id);
        };
```

Seen here: https://github.com/kbacha/stellar-sdk/blob/master/cli/src/trades.rs#L17

I believe this may be the last PR for the `cli` 🎉


cli/src/find_path.rs
```diff
@@ -15,19 +17,20 @@ pub fn find_path(client: &Client, matches: &ArgMatches) -> Result<()> {
...
-        let destination_amount: Amount = matches
+        let destination_amount = matches
             .value_of("amount")
             .expect("Destination amount is a required field")
-            .parse()
-            .expect("Amount must be properly formatted");
-        let destination_asset_str = matches
+            .parse::<Amount>()
+            .map_err(|_| String::from("Amount must be a valid i64 integer"))?;
+        let destination_asset_str: AssetIdentifier = matches
```
kbacha: This isn't a string any more after it's been parsed 😉 

you may want to remove the `_str` from the name.

cli/src/trades.rs
```diff
@@ -23,8 +23,12 @@ pub fn all(client: &Client, matches: &ArgMatches) -> Result<()> {
         if let (Some(base_str), Some(counter_str)) =
             (matches.value_of("base"), matches.value_of("counter"))
         {
-            let base = asset_identifier::from_str(base_str)?;
-            let counter = asset_identifier::from_str(counter_str)?;
+            let base: AssetIdentifier = base_str
+                .parse::<AssetIdentifier>()
+                .map_err(|_| String::from("Base asset must be properly formatted asset"))?;
+            let counter: AssetIdentifier = counter_str
```
kbacha: In some of these places you have the type specified more than once. You should only need to add the type either next to the variable `let counter: AssetIdentifier` or as part of the parse function `.parse::<AssetIdentifier>`. You can safely remove one.

robertDurst: Going with the second to keep everything consistent

cli/src/find_path.rs
robertDurst: Valid point

kbacha:
Thanks again!

PR #225, opened 2018-06-22
Cargo fmt update

kbacha:
clooney

ps: I think the new format is a bit nicer 😄


cli/src/assets.rs
```diff
@@ -2,8 +2,9 @@ use super::{cursor, ordering, pager::Pager};
 use clap::ArgMatches;
 use error::Result;
 use fmt::{Formatter, Simple};
-use stellar_client::{endpoint::asset,
-                     sync::{self, Client}};
```
robertDurst: Are you saying this isn't aesthetic :joy:

correlator:
🚀

PR #226, opened 2018-08-17
Fixes tiny error in docucomments regarding the horizon() fns

nebkor:
Hi there! I came across this project and was browsing the source when I noticed that the documentation strings for the non-test connection fns were erroneous. This tiny PR fixes that.


kbacha:
Thank you!

nebkor:
My pleasure; thanks for writing the client!

PR #228, opened 2018-08-20
Sometimes things need fixin

kbacha:
burgertime

Goes through and resolves a couple of failing specs. There is still one more failing one with the timeout. Will be looking at that soon.

also fixes cargo fmt.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -455,7 +455,8 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
                                 .long("resolution")
                                 .takes_value(true)
                                 .required(true)
-                                .help("Segment duration in format <number><unit> where units are s, m, h, d.  ie: 10h == 10 hours"),
+                                .possible_values(&["1m", "5m", "15m", "1h", "1d", "1w"])
+                                .help("The resolution of each segment. Limited to a set of valid inputs")
                         )
                         .arg(
                             Arg::with_name("base")
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "stellar-client"
-version = "0.1.1"
+version = "0.1.2"
 authors = ["Kevin Bacha <chewbacha@gmail.com>"]
 
 readme = "../README.md"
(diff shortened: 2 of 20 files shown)


client/src/endpoint/trade.rs
```diff
@@ -382,6 +382,36 @@ pub struct Aggregations {
...
+impl From<SegmentResolution> for u64 {
+    fn from(res: SegmentResolution) -> u64 {
+        match res {
+            SegmentResolution::OneMin => 60000,
+            SegmentResolution::FiveMin => 300000,
+            SegmentResolution::FifteenMin => 900000,
+            SegmentResolution::OneHour => 3600000,
+            SegmentResolution::OneDay => 86400000,
+            SegmentResolution::OneWeek => 604800000,
+        }
```
correlator: 🔥

client/src/endpoint/trade.rs
```diff
@@ -382,6 +382,36 @@ pub struct Aggregations {
...
+    OneMin,
+    /// Represents a five minute segment.
+    FiveMin,
+    /// Represents a fifteen minute segment.
+    FifteenMin,
+    /// Represents an hour long segment.
+    OneHour,
+    /// Represents a day long segment.
+    OneDay,
+    /// Respresents a week long segment.
```
correlator: These comments feel a bit superfluous.

cli/src/main.rs
```diff
@@ -455,7 +455,8 @@ fn build_app<'a, 'b>() -> App<'a, 'b> {
                                 .long("resolution")
                                 .takes_value(true)
                                 .required(true)
-                                .help("Segment duration in format <number><unit> where units are s, m, h, d.  ie: 10h == 10 hours"),
+                                .possible_values(&["1m", "5m", "15m", "1h", "1d", "1w"])
+                                .help("The resolution of each segment. Limited to a set of valid inputs")
```
correlator: very cool

client/src/endpoint/trade.rs
kbacha: Yes, but required for `#[deny(missing_doc)]`

client/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "stellar-client"
-version = "0.1.1"
+version = "0.1.2"
```
correlator: 💪

